PR #7, opened 2018-05-24
Magento v1.x SOAP Customer and Order export

ariens-shopify:
### What it does and why:
Provides sample scripts for exporting Magento v1.x customers and orders through the SOAP API.
### Checklist:

---

### Related issues:
Closes:

### DIFF
--- a/sample_scripts/third_party_platform_exporting/magento/v1.9/soap_api/README.md
+++ b/sample_scripts/third_party_platform_exporting/magento/v1.9/soap_api/README.md
@@ -27,7 +27,9 @@ These environment variables are required, or the script will exit early with an 
 
 These variables are not required but can be used to customize the extraction process:
 
-* `LAST_INCREMENT_ID`: The increment ID to resume extracting from in case the script is interrupted and needs to be restarted
+* `LAST_KEY_ID`: The last ID to resume extracting from in case the script is interrupted and needs to be restarted
+
+
 
 ### Setting environment variables
 
@@ -66,7 +68,7 @@ ruby customers.rb
 
 Depending on how many objects are being extracted, there may be an initial delay while the collection of objects is being fetched before they are displayed to the console output.
 
-For 50, 000 objects it could take up to 5-10 minutes before the JSON data starts to be displayed.
+For 50, 000 objects it could take up to 5-10 minutes before the script starts to print out it's status.  While the script is running, the key ID of each object will be printed to your console on stderr.  Once all the objects have been exported then the entirety of the JSON data will be printed to stdout.  See below about redirecting just the JSON data to a file.
 
 ## Redirecting into a file
 
@@ -80,7 +82,9 @@ Afterwards, inspect that file to ensure that it contains valid JSON data and tha
 
 ## Resuming
 
-In the event that your export is interrupted, you can resume from where you left off by defining the optional `LAST_INCREMENT_ID` environment variable.
+In the event that your export is interrupted, you can resume from where you left off by defining the optional `LAST_KEY_ID` environment variable.
+
+For orders, the key is `increment_id` and for customers the key is `customer_id`.  Examine the output or file created by the previous attempt and export the value in the `LAST_KEY_ID` environment variable before resuming and previously exported objects will be skipped.
 
 *NOTE*: These scripts can take several hours to run for larger shops and are not yet capable of making concurrent requests.  Future versions may address this by performing some API calls in parallel.
 
--- a/sample_scripts/third_party_platform_exporting/magento/v1.9/soap_api/README.md
+++ b/sample_scripts/third_party_platform_exporting/magento/v1.9/soap_api/README.md
@@ -2,7 +2,7 @@
 
 These are some sample scripts that demonstrate how objects can be exported from Magento.
 
-These command line tools look for specific environment variables that describe the Magento system being exported from then use the SOAP API to retrieve data and display JSON formatted data to the console/standard out.
+These command line tools look for specific environment variables that describe the Magento system being exported from. Next, they use the SOAP API to retrieve data and display JSON formatted data to the console/standard out.
 
 This output data is properly formatted to be supported as input into the Shopify Transporter Tool which then gets converted into CSV format used for uploading to the [Shopify Transporter App](https://apps.shopify.com/transporter).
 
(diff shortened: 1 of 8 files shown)


sample_scripts/third_party_platform_exporting/magento/v1.9/soap_api/README.md
```diff
@@ -0,0 +1,89 @@
+# Magento v1.x Export Scripts
+
+These are some sample scripts that demonstrate how objects can be exported from Magento.
+
+These command line tools look for specific environment variables that describe the Magento system being exported from. Next, they use the SOAP API to retrieve data and display JSON formatted data to the console/standard out.
```
JordanVerasamy: I think this section is a bit awkward. What about:

>These command line tools look for specific environment variables that describe the Magento system being exported from. Using that information, they use Magento's SOAP API to retrieve data from the store and output it in JSON format.

>These scripts output data in a format that the Shopify Transporter Tool can consume. The Shopify Transporter Tool can then transform that data into a CSV that can be uploaded to the [Shopify Transporter App](https://apps.shopify.com/transporter).

If that's not good enough, we should involve an actual content person, IMO.

sample_scripts/third_party_platform_exporting/magento/v1.9/soap_api/orders.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+require 'savon'
+require 'json'
+
+require_relative 'transporter_exporter'
+
+class OrderExporter < TransporterExporter
+  def run
+    orders_for(required_env_vars['MAGENTO_STORE_ID']).each do |order|
```
surajreddy: I'd prefer if this was done similar to `CustomerExporter` where the filtering of the store is done within the `customers` method, and we use `customers` here. Perhaps change the current `orders` method to `all_orders`?

sample_scripts/third_party_platform_exporting/magento/v1.9/soap_api/orders.rb
```diff
@@ -0,0 +1,47 @@
...
+      next if skip?(order)
+
+      puts JSON.pretty_generate(order.merge('items' => items_for(order)))
```
surajreddy: Do we need to add any rescues here and for customers above in case we fail to retrieve additional details for certain orders/customers?

sample_scripts/third_party_platform_exporting/magento/v1.9/soap_api/transporter_exporter.rb
```diff
@@ -0,0 +1,86 @@
...
+  def soap_client
+    @soap_client ||= Savon.client(
+      wsdl: "https://#{required_env_vars['MAGENTO_SOAP_API_HOSTNAME']}/api/v2_soap?wsdl",
+      open_timeout: 500,
+      read_timeout: 500,
+    )
+  end
+
+  def soap_session_id
+    @soap_session_id ||= soap_client.call(
```
surajreddy: 👍

sample_scripts/third_party_platform_exporting/magento/v1.9/soap_api/transporter_exporter.rb
```diff
@@ -0,0 +1,86 @@
...
+    MAGENTO_SOAP_API_HOSTNAME
+    MAGENTO_SOAP_API_USERNAME
+    MAGENTO_SOAP_API_KEY
+    MAGENTO_STORE_ID
+  ).freeze
+
+  OPTIONAL_ENV_VARS = %w(LAST_INCREMENT_ID).freeze
+
+  def soap_client
+    @soap_client ||= Savon.client(
```
surajreddy: 👍

sample_scripts/third_party_platform_exporting/magento/v1.9/soap_api/transporter_exporter.rb
```diff
@@ -0,0 +1,86 @@
...
+        username: required_env_vars['MAGENTO_SOAP_API_USERNAME'],
+        api_key: required_env_vars['MAGENTO_SOAP_API_KEY'],
+      }
+    ).body[:login_response][:login_return]
+  end
+
+  def skip?(item)
+    return false if optional_env_vars['LAST_INCREMENT_ID'].nil? || item[:incremenet_id].nil?
+
+    item[:increment_id].to_i < optional_env_vars['LAST_INCREMENT_ID'].to_i
```
surajreddy: I'm assuming this attribute is named the same for customers and orders.

sample_scripts/third_party_platform_exporting/magento/v1.9/soap_api/transporter_exporter.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  def required_env_vars
+    @required_env_vars ||= env_vars(REQUIRED_ENV_VARS, true)
+  end
+
+  def optional_env_vars
+    @optional_env_vars ||= env_vars(OPTIONAL_ENV_VARS, false)
+  end
+
+  def env_vars(vars, required = false)
```
surajreddy: 👍 nice!

sample_scripts/third_party_platform_exporting/magento/v1.9/soap_api/orders.rb
ariens-shopify: It's pretty similar to how you would call `.variants` on a `ShopifyAPI::Product` object.  There's no connection state across requests, so I don't foresee how it could blow up unless the product was maybe deleted during the execution of the export...?

sample_scripts/third_party_platform_exporting/magento/v1.9/soap_api/orders.rb
ariens-shopify: Yeah--I like that, the `MAGENTO_STORE_ID` is required, so there shouldn't be more than a single method.

sample_scripts/third_party_platform_exporting/magento/v1.9/soap_api/orders.rb
surajreddy: @ariens-shopify yes.

I'm worried about the soap call for these causing an exception that could fail the whole process/ we won't be able to recover from issues retrieving the rest of the orders if a single one fails. 

However, I'm content with adding an issue for this and dealing with this in the future - we don't have a recovery strategy if retrieving an order fails (ex: do we store the ids of all the orders and output them at the end, or do we retry, etc.)

Issue #14, opened 2018-06-05
Add a command line command to extract data from Magento


surajreddy:
Add an `extract` command to the gem that does the following:

Reads these parameters from the config file (or as passed to the command):
- Magento Hostname
- SOAP API Username
- SOAP API Key
- Start Date
- End Date
- Output file
- Type of object (orders/customers/products)

This should incorporate the requirements from #12 and #13:
- return a file that can be used for the convert step of the gem.
- should be resumable.
- accepts a start and end date for the data.

jennamsw:
- Create Gem integrated, tested extraction scripts 
- Pagination on updated scriptes

surajreddy:
This is now part of the newest release: https://github.com/Shopify/shopify_transporter/releases

PR #17, opened 2018-06-19
Integrate Magento 1.x SOAP API export scripts into CLI tool

JordanVerasamy:
### What it does and why:

Implements an `export` command, which can extract information from a Magento store instance and write to a file on disk. This file can then be used as input to the `convert` command.

The `export` command reads information about the Magento instance from the `config.yml` configuration file and then begins downloading information from Magento.

### Checklist:

---

### Related issues:
Closes:

https://github.com/Shopify/shopify_transporter/issues/14


exe/shopify_transporter
```diff
@@ -29,15 +30,41 @@ module ShopifyTransporter
...
+    method_option :object, type: :string, required: true,
+      desc: "The object type (customer, product, or order) to export.",
+      enum: %w(customer order)
+
+    desc 'export', 'Exports objects from a third-party platform into a format compatible for conversion.'
+    def export
+      api_key = $stdin.noecho do
+        ask('Please enter your Magento SOAP API key:')
+      end
+      say("\n")
```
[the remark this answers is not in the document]
JordanVerasamy: no reason, should prob just be `puts` -- since AFAIK `say` just passes along to `puts`

lib/shopify_transporter/exporter/magento/magento_customer_exporter.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    def initialize(store_id, client)
+      @client = client
+      @store_id = store_id
+    end
+
+    def export
+      puts "starting export..."
+      base_customers.map do |customer|
+        $stderr.puts "fetching customer: #{customer[:customer_id]}..."
```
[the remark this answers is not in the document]
JordanVerasamy: yeah, just a remnant of when we actually relied on stdout to be pure valid json. now that we're writing directly to a file, this should be on stdout

exe/shopify_transporter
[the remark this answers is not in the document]
JordanVerasamy: tbh, this was intended to be for debugging but we missed it on our cleanup pass over the branch

lib/shopify_transporter/exporter/magento/magento_exporter.rb
```diff
@@ -0,0 +1,19 @@
+require_relative './magento_customer_exporter.rb'
+require_relative './magento_order_exporter.rb'
+
+module ShopifyTransporter
+  class MagentoExporter
+    class MissingMagentoExporterError < ExportError; end
+
+    def self.for(store_id, type, client)
```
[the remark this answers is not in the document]
ariens-shopify: I'd love to have a deeper convo about this factory and ways that we can improve it.  All thoughts/feedback welcome!

lib/shopify_transporter/exporters/magento.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+require_relative 'magento/magento_exporter.rb'
```
mimperatore: It's probably ok to call this file/class `exporter`/`Exporter`, given that it's already in a `magento` directory.

lib/shopify_transporter/exporters/magento/customer_exporter.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+module ShopifyTransporter
+  module Exporters
+    module Magento
+      class CustomerExporter
+        attr_accessor :client, :store_id
```
mimperatore: Do we want the class user to be able to modify these after they're set by the initializer?  If not, maybe change to `attr_reader`.

mimperatore: Or maybe we don't even need `attr_` anything.

lib/shopify_transporter/exporters/magento/magento_exporter.rb
```diff
@@ -0,0 +1,25 @@
...
+
+      class MagentoExporter
+        def self.for(type, store_id, client)
+          case type
+          when 'customer'
+            CustomerExporter
+          when 'order'
+            OrderExporter
+          else
+            raise MissingExporterError
```
mimperatore: We don't seem to be rescuing this anywhere.

lib/shopify_transporter/exporters/magento/soap.rb
```diff
@@ -0,0 +1,47 @@
...
+          )
+        end
+
+        private
+
+        attr_accessor :hostname, :username, :api_key
+
+        def soap_client
+          @soap_client ||= Savon.client(
+            wsdl: "https://#{hostname}/api/v2_soap?wsdl",
```
mimperatore: Wonder if this would be useful as a config file setting.

lib/shopify_transporter/exporters/magento/soap.rb
```diff
@@ -0,0 +1,47 @@
...
+            open_timeout: 500,
```
mimperatore: Perhaps the timeouts as well.

lib/shopify_transporter/exporters/magento/soap.rb
```diff
@@ -0,0 +1,47 @@
...
+        def call(fn, params)
+          soap_client.call(
+            fn,
+            message: { session_id: soap_session_id }.merge(params)
+          )
+        end
+
+        private
+
+        attr_accessor :hostname, :username, :api_key
```
mimperatore: Do we need these to be accessible externally?

spec/shopify_transporter/exporters/magento/customer_exporter_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+        {
+          customer_id: 654321,
+          top_level_attribute: "an_attribute",
+          address_list: {
+            customer_address_attribute: "another_attribute",
+          }
+        }
+      ]
+
+      exporter = ShopifyTransporter::Exporters::Magento::CustomerExporter.new(1, soap_client)
```
mimperatore: What do you think of the idea of making the initializer parameter named?  Would make it easier to determine what `1` is.

spec/shopify_transporter/shopify_transporter_spec.rb
```diff
@@ -1,9 +1,11 @@
 # frozen_string_literal: true
 require 'shopify_transporter/pipeline/stage'
 RSpec.describe ShopifyTransporter do
-  def tmpfile(content)
+  def tmpfile(content, ext = nil)
```
mimperatore: This method's behaviour is a bit awkward.  Maybe just having it take the `content` and `extension`, and having it simply write it to a temp file would be sufficient, with the caller deciding what the format of the content should look like.

spec/shopify_transporter/shopify_transporter_spec.rb
```diff
@@ -306,6 +329,94 @@ def tmpfile(content)
     end
   end
 
+  describe ShopifyTransporter::Exporters::Exporter do
```
mimperatore: Any reason why this is not in its own test file?

spec/shopify_transporter/exporters/magento/customer_exporter_spec.rb
```diff
@@ -0,0 +1,61 @@
+# frozen_string_literal: true
+require 'shopify_transporter/pipeline/stage'
+
+RSpec.describe ShopifyTransporter::Exporters::Magento::CustomerExporter do
```
mimperatore: Alternatively, we could do:
```
module ShopifyTransporter
  module Exporters
     module Magento
        RSpec.describe CustomerExporter
```

spec/shopify_transporter/exporters/magento/customer_exporter_spec.rb
mimperatore: Can use `described_class` here.

spec/shopify_transporter/shopify_transporter_spec.rb
```diff
@@ -306,6 +329,94 @@ def tmpfile(content)
...
+    class SomePlatformExporter
+      def export
+        [{ foo: 'bar' }]
+      end
+    end
+
+    it 'writes exported data to file' do
```
mimperatore: Add `describe '#run' do`... as I think all these tests are describing the `run` method.

spec/shopify_transporter/shopify_transporter_spec.rb
```diff
@@ -306,6 +329,94 @@ def tmpfile(content)
...
+      in_temp_folder do
```
mimperatore: Not sure why we need this.  Can we get rid of it?

exe/shopify_transporter
```diff
@@ -29,15 +30,35 @@ module ShopifyTransporter
       conversion_error(error.message)
     end
 
+    method_option :config, type: :string, default: 'config.yml',
+      desc: "The config file that describes the third party platform to export from."
```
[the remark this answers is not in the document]
JordanVerasamy: Yeah, Rubocop doesn't complain about that. Wonder if we have it configured differently between this project and the Transporter app. @fahia do you know how to check this?

lib/shopify_transporter/exporters/magento.rb
JordanVerasamy: I was just a bit worried about ambiguous naming potentially leading to future issues, similar to the `RACC::ParseError` stuff we encountered in the App. Do we think that having identically named classes in nested modules is okay?

lib/shopify_transporter/exporters/magento/magento_exporter.rb
JordanVerasamy: I don't believe that _should_ be rescued anywhere.

Thor knows that it expects either `customer` or `order` as the object type. If it receives something else, that is handled by Thor:

```
$ exe/shopify_transporter export --object=bloop --config=config.yml
Expected '--object' to be one of customer, order; got bloop
```

It is truly an exceptional circumstance if an unexpected parameter gets through Thor's check and makes it to this function. So, I think it is reasonable for an exception to get bubbled up to the top level.

One change I would maybe accept is to just not even raise an error in this case, since it literally should never happen

spec/shopify_transporter/shopify_transporter_spec.rb
JordanVerasamy: This method is super super awkward. I was focusing on making it backwards compatible so I wouldn't have to change 100+ lines of test files in the same PR as an unrelated 500 lines of code change. The intent was for a future PR to refactor this into something more sane.

JordanVerasamy: Looking at it now, the changes to all those other tests wouldn't be _that_ intensive -- probably just a line or two change per test. Still, I'd prefer to do it in a separate PR since this is already so large. Thoughts?

spec/shopify_transporter/shopify_transporter_spec.rb
JordanVerasamy: Oh yeah, that's from when we were generating output files and the test expectations were placed on the file contents. Now that we are writing to standard out, this is not required.

spec/shopify_transporter/shopify_transporter_spec.rb
JordanVerasamy: Only because all the tests for other analogous commands are in this file and I didn't want to break away from the established pattern. Happy to move them out.

lib/shopify_transporter/exporters/magento/soap.rb
JordanVerasamy: I think this is scope creep and I would rather merge the PR with a more minimal functionality set. We can always add more things later if there is a need.

lib/shopify_transporter/exporters/magento/soap.rb
JordanVerasamy: Same as above.

.rubocop-https---shopify-github-io-ruby-style-guide-rubocop-yml
```diff
@@ -1,3 +1,5 @@
+# Recommended rubocop version: ~> 0.56.0
```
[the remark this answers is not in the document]
JordanVerasamy: Had a discussion w/ Suraj and Fahia about this. I feel confident that this is the right change.

spec/shopify_transporter/exporters/exporter_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+              "hostname" => 'magento.host',
+              "username" => 'something',
+              "api_key" => 'a_key',
+            },
+            "store_id" => 1,
+          },
+        }
+      end
+
+      it 'writes exported data to file' do
```
surajreddy: Suggest rename to `writes exported data to stdout`

spec/shopify_transporter/exporters/exporter_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+        config_filename = 'nonexistent_config.yml'
+
+        error_message = "Invalid configuration: cannot find file name 'nonexistent_config.yml'"
+
+        expect { Exporter.new(config_filename, :unused) }
+          .to raise_error(InvalidConfigError, error_message)
+      end
+
+      it 'raises InvalidConfigError if config file is missing username' do
+        config_without_username = default_config.tap { |cfg| cfg['export_configuration']['soap'].delete('username') }
```
surajreddy: 😎 nice use of tap!

lib/shopify_transporter/exporters/magento/magento_exporter.rb
surajreddy: I'm okay with raising and not rescuing here - we do want this to blow up if an invalid option got this far, and this gives us a specific error when debugging/solving it.

exe/shopify_transporter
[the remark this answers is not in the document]
fahia: From `rubocop.yml`:
```
Style/StringLiteralsInInterpolation:
  EnforcedStyle: single_quotes
  SupportedStyles:
  - single_quotes
  - double_quotes
```
Is the culprit having `double_quotes` in `SupportedStyles` ?

lib/shopify_transporter/exporters/magento/order_exporter.rb
```diff
@@ -0,0 +1,46 @@
...
+  module Exporters
+    module Magento
+      class OrderExporter
+        def initialize(store_id: nil, client: nil)
+          @client = client
+          @store_id = store_id
+        end
+
+        def export
+          $stderr.puts "Starting export..."
```
fahia: Replace `$stderr` with `puts`, we are are not logging bad things and can simply log to stdout.

lib/shopify_transporter/exporters/magento/order_exporter.rb
```diff
@@ -0,0 +1,46 @@
...
+          base_orders.map do |order|
+            $stderr.puts "Fetching order: #{order[:increment_id]}..."
```
fahia: Same with here re: puts

spec/shopify_transporter/exporters/magento/customer_exporter_spec.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+require 'shopify_transporter/pipeline/stage'
+
+module ShopifyTransporter
+  module Exporters
+    module Magento
+      RSpec.describe CustomerExporter do
+        context '#run' do
```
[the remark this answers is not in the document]
JordanVerasamy: empty hash.

exe/shopify_transporter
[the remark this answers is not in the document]
JordanVerasamy: @surajreddy that fixed it! :D

spec/shopify_transporter/shopify_transporter_spec.rb
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 require 'shopify_transporter/pipeline/stage'
+require 'shopify_transporter/pipeline/stage'
```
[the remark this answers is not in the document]
fahia: 🔥 🔥

PR #23, opened 2018-09-07
Extract product mappings from a magento database

surajreddy:
### What it does and why:
- This is a work in progress script to extract product mappings from a magento database.
- This will be integrated in the data extraction process in the future.

### Checklist:

### DIFF
--- a/lib/shopify_transporter.rb
+++ b/lib/shopify_transporter.rb
@@ -152,11 +152,11 @@ class TransporterTool
   def stage_class_from(name, type)
     class_name = stage_classname(name, type)
     klass = begin
-      k = Object.const_get(class_name)
-      k.is_a?(Class) && k
-    rescue NameError
-      nil
-    end
+              k = Object.const_get(class_name)
+              k.is_a?(Class) && k
+            rescue NameError
+              nil
+            end
     raise StageNotFoundError, name unless class_exists?(klass)
     klass
   end
--- a/shopify_transporter.gemspec
+++ b/shopify_transporter.gemspec
@@ -31,7 +31,9 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'rake', '~> 12.3'
 
   spec.add_dependency 'activesupport', '~> 5.1'
+  spec.add_dependency 'mysql2'
+  spec.add_dependency 'savon', '~> 2.12'
+  spec.add_dependency 'sequel', '~> 5.12'
   spec.add_dependency 'thor', '~> 0.20'
   spec.add_dependency 'yajl-ruby', '~> 1.3'
-  spec.add_dependency 'savon', '~> 2.12'
 end
--- a/lib/shopify_transporter/exporters/magento/product_exporter.rb
+++ b/lib/shopify_transporter/exporters/magento/product_exporter.rb
@@ -10,7 +10,7 @@ module ShopifyTransporter
         end
 
         def export
-          $stderr.puts "Starting export..."
+          $stderr.puts 'Starting export...'
           base_products.compact
         end
 
--- a/lib/shopify_transporter/exporters/magento/product_mapping_exporter.rb
+++ b/lib/shopify_transporter/exporters/magento/product_mapping_exporter.rb
@@ -1,0 +1,64 @@
+# frozen_string_literal: true
+require 'sequel'
+
+module ShopifyTransporter
+  module Exporters
+    module Magento
+      class ProductMappingExporter
+        BATCH_SIZE = 1000
+
+        def initialize(
+          database: '',
+          host: '',
+          port: 3306,
+          user: '',
+          password: '',
+          filename: 'magento_product_mappings.csv'
+        )
+          @host = host
+          @port = port
+          @user = user
+          @password = password
+          @filename = filename
+          @database = database
+        end
+
+        def extract_mappings
+          write_headers
+
+          Sequel.connect(
+            adapter: :mysql2, user: @user, password: @password, host: @host, port: @port, database: @database
+          ) do |db|
+            ordered_mappings = db
+              .from(:catalog_product_relation)
+              .order(:parent_id)
+            current_id = ordered_mappings.first[:parent_id]
+            max_id = ordered_mappings.last[:parent_id]
+
+            while current_id < max_id
+              mappings_batch = ordered_mappings.where(parent_id: current_id...(current_id + BATCH_SIZE))
+              write_data(mappings_batch)
+              current_id += BATCH_SIZE
+            end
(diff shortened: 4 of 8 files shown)


lib/wip/product_mapping_exporter.rb
```diff
@@ -0,0 +1,71 @@
+# frozen_string_literal: true
+
+require 'sequel'
+
+module Magento
+  class ProductMappingExporter
+    ADAPTERS = {
+      'mysql': :mysql2,
+      'postgres': :postgres,
```
fahia: We need to pull in the `pg` gem if we've specified we're using the adapter

fahia: We need to pull in the `pg` gem if we've specified we're using the adapter

lib/wip/product_mapping_exporter.rb
```diff
@@ -0,0 +1,71 @@
+# frozen_string_literal: true
```
mimperatore: Reminder: the `wip` directory should be 💣

lib/wip/product_mapping_exporter.rb
```diff
@@ -0,0 +1,71 @@
...
+    ADAPTERS = {
+      'mysql': :mysql2,
+      'postgres': :postgres,
+    }
+
+    BATCH_SIZE = 1000
+
+    def initialize(
+      database_type: :mysql,
+      database_name:,
```
mimperatore: Is there a default value for this?

lib/wip/product_mapping_exporter.rb
```diff
@@ -0,0 +1,71 @@
+# frozen_string_literal: true
+
+require 'sequel'
+
+module Magento
+  class ProductMappingExporter
```
mimperatore: Wondering if we should turn this into a thor subcommand.

lib/wip/product_mapping_exporter.rb
```diff
@@ -0,0 +1,71 @@
...
+        adapter: adapter, user: username, password: password, host: host, port: port, database: database_name
+      )
+
+      mappings = db.from(:catalog_product_relation)
+      ordered_mappings = mappings.order(:parent_id)
+      current_id = ordered_mappings.first[:parent_id]
+      max_id = ordered_mappings.last[:parent_id]
+
+      while current_id <= max_id do
+        puts "doing batch #{current_id}"
```
mimperatore: May want to use `STDERR` here, esp. if we move to converting this to a thor subcommand and the output could go to `STDOUT`, like the other subcommands.

lib/wip/product_mapping_exporter.rb
```diff
@@ -0,0 +1,71 @@
...
+        mappings_batch = ordered_mappings.where(parent_id: current_id..(current_id + BATCH_SIZE))
+        write_data(mappings_batch)
+        current_id += BATCH_SIZE
+      end
+    end
+
+    private
+
+    attr_accessor :host, :username, :password, :filename, :adapter, :port, :database_name
```
mimperatore: Do these need to be accessible outside of this class?

lib/wip/product_mapping_exporter.rb
```diff
@@ -0,0 +1,71 @@
...
+
+    def write_headers
+      File.open(filename, 'w') do |file|
+        file << "product_id,associated_product_id\n"
```
mimperatore: Maybe use `$/`?

lib/wip/product_mapping_exporter.rb
```diff
@@ -0,0 +1,71 @@
...
+
+      @adapter = ADAPTERS[database_type]
+      @host = host
+      @port = port
+      @username = username
+      @password = password
+      @filename = filename
+      @database_name = database_name
+
+      write_headers
```
mimperatore: Suggest moving this to `extract_mappings`.

lib/wip/product_mapping_exporter.rb
```diff
@@ -0,0 +1,71 @@
...
+    end
+
+    def extract_mappings
+      db = Sequel.connect(
```
mimperatore: May need to close the connection in an `ensure` block.

lib/wip/product_mapping_exporter.rb
JordanVerasamy: Would you ever want to run this without also running `export`? I'm thinking maybe this can be called from other code that is part of a Thor subcommand, and the user never has to run this themselves. But maybe (for example if it is long-running) it is better for the user to be able to run them separately.

fahia: I agree that this doesn't make sense to run in isolation, It makes sense that there'd be a single public subcommand that is called that is calling this exporter somewhere internally.

lib/wip/product_mapping_exporter.rb
surajreddy: woops, this was added just for debugging purposes, I forgot to remove it.

shopify_transporter.gemspec
```diff
@@ -33,4 +33,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'activesupport', '~> 5.1'
   spec.add_dependency 'thor', '~> 0.20'
   spec.add_dependency 'yajl-ruby', '~> 1.3'
+  spec.add_dependency 'sequel', '~> 5.12'
+  spec.add_dependency 'mysql2'
+  spec.add_dependency 'pg'
```
surajreddy: I think we can stick with just mysql: https://magento.stackexchange.com/questions/30969/magento-using-postgres

surajreddy:
This is now good for 👀

lib/shopify_transporter.rb
```diff
@@ -152,11 +152,11 @@ def custom_stage_class_from(stage_name)
   def stage_class_from(name, type)
     class_name = stage_classname(name, type)
     klass = begin
-      k = Object.const_get(class_name)
-      k.is_a?(Class) && k
-    rescue NameError
-      nil
-    end
+              k = Object.const_get(class_name)
```
surajreddy: new rubocop rules!

surajreddy:
**Code snippet to test the script:**
```
load 'lib/shopify_transporter/exporters/magento/product_mapping_exporter.rb'

exporter = ShopifyTransporter::Exporters::Magento::ProductMappingExporter.new(
  database: '...',
  host: '...',
  user: '...',
)

exporter.extract_mappings
```

JordanVerasamy:
sick tests

JordanVerasamy:
wonder what's up with Travis CI though.

PR #31, opened 2018-09-12
Remove filters from order and customer soap calls to retrieve all data

surajreddy:
### What it does and why:
- Removes the `store_id` filter for orders and customers, because this limits the data returned.

### Checklist:

---

### DIFF
--- a/lib/shopify_transporter/exporters/magento/customer_exporter.rb
+++ b/lib/shopify_transporter/exporters/magento/customer_exporter.rb
@@ -4,9 +4,8 @@ module ShopifyTransporter
   module Exporters
     module Magento
       class CustomerExporter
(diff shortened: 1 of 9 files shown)


lib/shopify_transporter/exporters/magento/customer_exporter.rb
```diff
@@ -21,24 +20,13 @@ def export
         private
 
         def base_customers
-          result = @client.call(:customer_customer_list, filters: filters).body
+          result = @client.call(:customer_customer_list, filters: nil).body
```
JordanVerasamy: nit: you could just pass `{}` as the params hash to the `call` method

JordanVerasamy: not blocking at all

PR #34, opened 2018-09-14
push code to fix line item tax not being exported

zliu1020:
### What it does and why:
- Fixed the issue where line item taxes are not being exported properly. 
- Updated the corresponding tests.
- Added IDE related folder to .gitignore

### Checklist:

---

### Related issues: N/A
Closes: N/A

### DIFF
--- a/lib/shopify_transporter/pipeline/magento/order/line_items.rb
+++ b/lib/shopify_transporter/pipeline/magento/order/line_items.rb
@@ -42,11 +42,12 @@ module ShopifyTransporter
           end
 
           def tax_lines(item)
-            {
+            # Magento supports only one type of tax, the tax_lines should always be in array format
+            [{
               title: 'Tax',
               price: item['tax_amount'],
               rate: item['tax_percent'],
-            }.stringify_keys
+            }.stringify_keys]
           end
         end
       end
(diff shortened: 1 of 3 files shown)


lib/shopify_transporter/pipeline/magento/order/line_items.rb
```diff
@@ -42,11 +42,12 @@ def line_item(item)
           end
 
           def tax_lines(item)
-            {
+            # Magento supports only one type of tax, the tax_lines should always be in array format
+            [{
```
surajreddy: This is minor, if our linter did not catch it then don't worry about changing it. I prefer if we split out the `[ &  {` onto separate lines here and in the test:
```
  [
    {
      ...
    }
  ]
```

fahia: Do we need the comment btw? I think this was ultimately confusion on our end but may be obvious to others. I'd vote for leaving the comment out.

surajreddy: 👍 to removing the comment - in general we try to avoid comments except for exceptional circumstances. In this case the test documents the expected format as well.

PR #36, opened 2018-09-17
some minor clean up and tweaking

fahia:
### What it does and why:
- Cleans up some of the extraction code with some renaming, and refactoring
- Re-uses the sql connection so that we're not hammering the mysql instance
### Checklist:

---

### Related issues:
Closes:


lib/shopify_transporter/exporters/exporter.rb
```diff
@@ -68,9 +69,9 @@ def ensure_config_has_required_keys
         ]
 
         product_required_keys = [
-          %w(export_configuration database hostname),
-          %w(export_configuration database username),
-          %w(export_configuration database name),
+          %w(export_configuration database host),
+          %w(export_configuration database user),
+          %w(export_configuration database database),
```
surajreddy: Minor: I prefer leaving the last one as `name`, or maybe `database_name` because having a nested `database` key for the `database` hash seems odd.

fahia: Fair, I just wanted to map it directly to the values a database expects (specifically mysql). It's been way less confusing that way, IME.

JordanVerasamy: i had this exact conversation in my head when i wrote this, i dislike `database database` as well personally. but i really dont care much

PR #37, opened 2018-09-18
Disconnect database properly after yielding

JordanVerasamy:
### What it does and why:

Makes sure to disconnect from the database after the given block is run. This may be the reason we were encountering weird unpredictable behaviour with database connection limits last week.

### Checklist:

---

### Related issues:
Closes:

### DIFF
--- a/lib/shopify_transporter/exporters/magento/sql.rb
+++ b/lib/shopify_transporter/exporters/magento/sql.rb
@@ -30,6 +30,7 @@ module ShopifyTransporter
             database: @database
           )
           yield(@connection)
+        ensure
           @connection.disconnect
         end
       end
--- a/lib/shopify_transporter/exporters/magento/sql.rb
+++ b/lib/shopify_transporter/exporters/magento/sql.rb
@@ -30,6 +30,7 @@ module ShopifyTransporter
             database: @database
           )
           yield(@connection)
+          @connection.disconnect
         end
       end
     end
--- a/lib/shopify_transporter/exporters/magento/sql.rb
+++ b/lib/shopify_transporter/exporters/magento/sql.rb
@@ -21,17 +21,16 @@ module ShopifyTransporter
         end
 
         def connect
-          @connection ||= Sequel.connect(
+          Sequel.connect(
             adapter: :mysql2,
             user: @user,
             password: @password,
             host: @host,
             port: @port,
             database: @database
-          )
-          yield(@connection)
-        ensure
-          @connection.disconnect
+          ) do |connection|
+            yield connection
+          end
         end
       end
     end
--- a/spec/files/config-with-all-platforms-stage-type.yml
+++ b/spec/files/config-with-all-platforms-stage-type.yml
@@ -1,0 +1,8 @@
+platform_type: Magento
+object_types:
+  customer:
+    record_key: email
+    key_required: true
+    pipeline_stages:
+      - TopLevelAttributes:
+        type: all_platforms
--- a/spec/shopify_transporter/exporters/magento/sql_spec.rb
+++ b/spec/shopify_transporter/exporters/magento/sql_spec.rb
@@ -28,12 +28,13 @@ module ShopifyTransporter
               user: 'dbuser',
               password: 'some_password'
             ).and_return(connection)
-            expect(connection).to receive(:disconnect)
 
             sql_client.connect { |db| nil }
           end
 
-          it 'disconnects the database even if the given block raises an error' do
+          it 'yields the db connection on SQL#connect' do
+            connection = double('db connection')
+
             expect(Sequel).to receive(:connect).with(
               adapter: :mysql2,
               database: 'magento',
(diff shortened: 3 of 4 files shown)


lib/shopify_transporter/exporters/magento/sql.rb
```diff
@@ -30,6 +30,8 @@ def connect
             database: @database
           )
           yield(@connection)
+        ensure
+          @connection.disconnect
```
fahia: Not sure if this is necessary if we're yielding inside the connection block because:
image

JordanVerasamy: Yeah. Totally possible that I am not understanding how this works. Here's my current understanding and why I thought this was necessary:

I'm not providing a block to `Sequel#connect`. I'm calling `Sequel#connect` and then yielding the return value of that call. That means that the code path I'm following inside Sequel is not the one that consumes a block, but rather the one that just returns a connection that is ready for use. Then when I yield that connection to the block given to `SQL#connect`, there isn't any code that disconnects that connection afterwards.

There may be a better way to accomplish this than using `ensure`. For example, what if when `SQL#connect` received a block, it directly passed the block to `Sequel#connect`?

fahia: Yeah, I don't think we're actually yielding within a block so we don't actually get the disconnect for free.

How about this?
```ruby
          @connection ||= Sequel.connect(
            adapter: :mysql2,
            user: @user,
            password: @password,
            host: @host,
            port: @port,
            database: @database
          ) do |db|
            yield db
          end
```

I'm not sure how to check the connection closes but that does yield within the block. I _assume_ this would disconnect once we exit the block.

JordanVerasamy: Yeah, I think that would work.

fahia: Hahaha I re-read it just now and realized we are in agreement 😄

spec/files/config-with-all-platforms-stage-type.yml
```diff
@@ -0,0 +1,8 @@
+platform_type: Magento
```
surajreddy: Is this file being used anywhere?

spec/shopify_transporter/exporters/magento/sql_spec.rb
```diff
@@ -6,23 +6,42 @@ module Exporters
...
               host: 'magento-instance.domain.com',
               port: 1234,
               user: 'dbuser',
               password: 'some_password'
-            )
+            ).and_return(connection)
+
+            sql_client.connect { |db| nil }
+          end
+
```
surajreddy: We should add a test that it yields the connection.

lib/shopify_transporter/exporters/magento/sql.rb
```diff
@@ -28,8 +28,9 @@ def connect
             host: @host,
             port: @port,
             database: @database
-          )
-          yield(@connection)
+          ) do |db|
```
surajreddy: Suggestion: reword `db` to `connection`

lib/shopify_transporter/exporters/magento/sql.rb
```diff
@@ -28,8 +28,9 @@ def connect
...
+            yield db
+          end
```
surajreddy: We're not reusing `@connection` anywhere, I suggest removing the assignment to the instance variable.

Maybe rename the method to `with_connection` to match how we yield objects like this in other projects?

spec/shopify_transporter/exporters/magento/sql_spec.rb
```diff
@@ -6,23 +6,42 @@ module Exporters
...
+          it 'disconnects the database even if the given block raises an error' do
+            expect(Sequel).to receive(:connect).with(
+              adapter: :mysql2,
+              database: 'magento',
+              host: 'magento-instance.domain.com',
+              port: 1234,
+              user: 'dbuser',
+              password: 'some_password'
+            ).and_yield(connection)
+
```
surajreddy: This isn't raising an error right now.

spec/files/config-with-all-platforms-stage-type.yml
fahia: Yes, in a test in `shopify_transporter_spec`

spec/shopify_transporter/exporters/magento/sql_spec.rb
fahia: a remnant of the last test, shouldn't raise an error, changed the title.

lib/shopify_transporter/exporters/magento/sql.rb
fahia: The underlying sequel lib yields a connection in `Sequel#connect`, which was the rationale behind keeping it consistent in the wrapper. I don't have a preference between `with_` `connection`, though

PR #39, opened 2018-09-19
Map top level Magento product attributes to shopify products; Map simple products as variants to configurable products

fahia:
### What it does and why:
- Map top level Magento product attributes to shopify products
- Map simple products as variants to configurable products
### Checklist:


lib/shopify_transporter/record_builder/product_record_builder.rb
```diff
@@ -0,0 +1,32 @@
...
+    private
+
+    def has_parent?(product)
+      product['parent_id'].present?
+    end
+
+    def parent_record(product)
+      parent_id = product['parent_id']
+      record = @instances[parent_id] ||= {}
+      @last_record = record
```
JordanVerasamy: what is this used for again?

lib/shopify_transporter.rb
```diff
@@ -235,9 +235,15 @@ def valid_file?(path)
 
   def build_classes_based_on_config
     @record_class = Object.const_get("ShopifyTransporter::Shopify::#{@object_type.capitalize}")
-    @record_builder = ShopifyTransporter::RecordBuilder.new(
-      record_key_from_config, key_required_from_config
-    )
+    @record_builder = if @object_type.capitalize == 'Product'
```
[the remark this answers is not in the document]
surajreddy: 👍 to Jordan's suggestion.

lib/shopify_transporter/exporters/magento/product_exporter.rb
```diff
@@ -17,7 +17,7 @@ def export
           $stderr.puts 'Starting export...'
           products = base_products.map do |product|
             $stderr.puts "Fetching product: #{product[:product_id]}"
-            product.merge(items: info_for(product[:product_id]))
+            product.merge(info_for(product[:product_id]))
```
surajreddy: 👍

lib/shopify_transporter/pipeline/magento/product/top_level_attributes.rb
```diff
@@ -14,19 +14,28 @@ def convert(hash, record)
 
           class TopLevelAttributesAccumulator < Shopify::AttributesAccumulator
             COLUMN_MAPPING = {
-              'sku' => 'sku',
               'name' => 'title',
               'description' => 'body_html',
+              'url_key' => 'handle',
+              'created_at' => 'created_at',
```
surajreddy: We don't import `created_at`, so let's remove this field.

lib/shopify_transporter/pipeline/magento/product/top_level_attributes.rb
```diff
@@ -14,19 +14,28 @@ def convert(hash, record)
...
-              map_from_key_to_val(COLUMN_MAPPING, input)
+              attributes = map_from_key_to_val(COLUMN_MAPPING, input)
+              attributes['published_scope'] = published_scope(input)
+              attributes
+            end
+
+            def published_scope(input)
+              value = input['visibility']
+              return unless value.present?
+              value == '1' ? 'web' : ' '
```
surajreddy: This should check that `visibility != nil && visibility !=1` instead.

We should also add these fields if `visibility != nil && visibility !=1` :
```
Published
Published At (can be set to Updated At)
```

lib/shopify_transporter/pipeline/magento/product/variant_attributes.rb
```diff
@@ -0,0 +1,28 @@
...
+          def convert(hash, record)
+            accumulator = VariantAttributesAccumulator.new(record)
+            accumulator.accumulate(hash)
+          end
+
+          class VariantAttributesAccumulator < Shopify::AttributesAccumulator
+            def accumulate(current_product)
+              if current_product['parent_id'].present?
+                @output['variants'] = [] if @output['variants'].nil?
+                @output['variants'].push(current_product)
```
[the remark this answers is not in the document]
surajreddy: 👍 to using `<<` instead of push, but agree that this is minor.

lib/shopify_transporter/record_builder/product_record_builder.rb
surajreddy: This is used if the row does not have a key or a parent_id. The record builder returns the last processed object.

This supports data like this:
```
Name, Line Item 1
ORD1, Banana
, Apple
```

(It will tie Apple to the last processed Order, ORD1)

lib/shopify_transporter.rb
```diff
@@ -235,7 +235,12 @@ def valid_file?(path)
 
   def build_classes_based_on_config
     @record_class = Object.const_get("ShopifyTransporter::Shopify::#{@object_type.capitalize}")
-    @record_builder = ShopifyTransporter::RecordBuilder.new(
+    record_builder_class = if @object_type.capitalize == 'Product'
```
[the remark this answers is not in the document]
JordanVerasamy: I think this is the best option in terms of complexity versus value:

>Perhaps there needs to be a record builder for each object type (where they all just call super except product) 

since you already have the `@record_class` here.

I agree that this isn't optimal but I'd be ok with it going in for now since we are on a time limit and this isn't so complex that it _requires_ an abstraction.

JordanVerasamy:
This closes https://github.com/Shopify/transporter_app/issues/1313 and https://github.com/Shopify/transporter_app/issues/1289.

PR #40, opened 2018-09-19
Require all database fields and make key missing error clearer

surajreddy:
### What it does and why:
- Requires all database connection keys.
- Makes the error message for missing keys more descriptive to be able to locate the right keys.

### Checklist:

### DIFF
--- a/lib/shopify_transporter/exporters/exporter.rb
+++ b/lib/shopify_transporter/exporters/exporter.rb
@@ -70,14 +70,16 @@ module ShopifyTransporter
 
         product_required_keys = [
           %w(export_configuration database host),
-          %w(export_configuration database user),
+          %w(export_configuration database port),
           %w(export_configuration database database),
+          %w(export_configuration database user),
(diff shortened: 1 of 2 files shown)


JordanVerasamy:
What's the purpose of requiring those database fields in the config? I'm guessing a lot of users will have their DB listening on the default Magento DB port 3306, and there's no point forcing them to include it in the config file AFAIK.

surajreddy:
@JordanVerasamy This makes it more explicit/it's easier to notice that it's an option. We can set it to `3306` in the `config.yml` file we generate if you like, or accept an empty value for it, but because the config is generated I think it's okay. This reduces the number of issues someone might run into and gives them a clear message.

Instead of debugging why the database connection is failing, they can catch this before it happens.

fahia:
I agree that generally, they'll be listening on the default port so having the config generate the value as the default would make sense (this also allows them to change it if they are not listening on the default port). However, I think we should still include that in the config because of the possibility of customized configurations.

PR #45, opened 2018-09-24
Extract and convert top level variant fields 

zliu1020:
### What it does and why:
Add a new pipeline stage to handle top level variant field conversion into Shopify format

### Checklist:

---

### Related issues: https://github.com/Shopify/transporter_app/issues/1283
Closes:


lib/shopify_transporter/pipeline/magento/product/top_level_variant_attributes.rb
```diff
@@ -0,0 +1,38 @@
...
+            }
+            def input_applies?(input)
+              true unless input['parent_id'].nil?
+            end
+
+            def attributes_from(input)
+              map_from_key_to_val(COLUMN_MAPPING, input)
+            end
+
+            def accumulate_attributes(attributes)
```
JordanVerasamy: Why do we need the `delete_if`? Just curious.

JordanVerasamy:
Hmm, tried to 🎩, unconvinced that it works as expected. Will play around with it a bit more.

lib/shopify_transporter/pipeline/magento/product/top_level_variant_attributes.rb
zliu1020: The function argument ```record``` comes in with this format: an key value pair with key being```variants``` and value being an array of simple products. The simple products are from the last pipeline stage so they are in Magento format. The delete_if is to get rid of the simple product in Magento format and the code in the next line will insert the product in Shopify format to the array

lib/shopify_transporter/pipeline/magento/product/top_level_variant_attributes.rb
```diff
@@ -0,0 +1,38 @@
...
+require 'shopify_transporter/pipeline/stage'
+require 'shopify_transporter/shopify'
+
+module ShopifyTransporter
+  module Pipeline
+    module Magento
+      module Product
+        class TopLevelVariantAttributes < Pipeline::Stage
+          def convert(hash, record)
+            accumulator = TopLevelVariantAttributesAccumulator.new(record['variants'])
```
surajreddy: This should also fetch the correct variant from `record['variants']` by checking that the id of variants matches hash.

spec/shopify_transporter/pipeline/magento/product/top_level_variant_attributes_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+        child_product = FactoryBot.build(:advanced_simple_product, with_nonsense)
+        parent_product = FactoryBot.build(:advanced_configurable_product, {variants: [child_product]})
+        variants_in_shopify_format = described_class.new.convert(child_product, parent_product)
+        expected_variants_in_shopify_format = [
+          {
+            sku: 'm100',
+            grams: '100',
+            price: '222'
+          }.deep_stringify_keys
+        ]
```
surajreddy: Can you add a test where a product has 2 variants? Does this extract the correct variant and update the right variant in this case?

lib/shopify_transporter/pipeline/magento/product/top_level_variant_attributes.rb
```diff
@@ -0,0 +1,38 @@
...
+              true unless input['parent_id'].nil?
+            end
+
+            def attributes_from(input)
+              map_from_key_to_val(COLUMN_MAPPING, input)
+            end
+
+            def accumulate_attributes(attributes)
+              @output.delete_if { |product| product['sku'] == attributes['sku'] }
+              @output << attributes
```
surajreddy: Offline conversation recap: I spoke with Ziyan about changing this behavior to choose the right variant before modifying it. This way we don't lose data from other stages, and the order of the pipeline stages can be changed.

lib/shopify_transporter/pipeline/magento/product/top_level_variant_attributes.rb
fahia: I think we don't actually care about them enough to delete them as there's no side effect of their presence after-the-fact. As long as the attributes are moved to the right hash, all is good

JordanVerasamy:
I can't seem to get this to actually convert the weight. Am I doing something wrong?

Here's some sample input and output:

Input:

https://pastebin.com/viiph4bJ

Then I run `exe/shopify_transporter convert --object=product products.json > shopify_products.csv`:

Output:

```
Handle,Title,Body (HTML),Vendor,Type,Tags,Template Suffix,Published Scope,Published,Published At,Option1 Name,Option1 Value,Option2 Name,Option2 Value,Option3 Name,Option3 Value,Variant SKU,Metafields Global Title Tag,Metafields Global Description Tag,Metafield Namespace,Metafield Key,Metafield Value,Metafield Value Type,Variant Grams,Variant Inventory Tracker,Variant Inventory Qty,Variant Inventory Policy,Variant Fulfillment Service,Variant Price,Variant Compare At Price,Variant Requires Shipping,Variant Taxable,Variant Barcode,Image Attachment,Image Src,Image Position,Image Alt Text,Variant Image,Variant Weight Unit,Variant Tax Code
french-cuff-cotton-twill-oxford,French Cuff Cotton Twill Oxford,"Button front. Long sleeves. Tapered collar, chest pocket, french cuffs.",,,,,global,true,2013-03-20 16:58:34,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
french-cuff-cotton-twill-oxford,,,,,,,,,,,,,,,,msj000,,,,,,,,,,,,190.0000,,,,,,,,,,,
french-cuff-cotton-twill-oxford,,,,,,,,,,,,,,,,msj001,,,,,,,,,,,,190.0000,,,,,,,,,,,
french-cuff-cotton-twill-oxford,,,,,,,,,,,,,,,,msj002,,,,,,,,,,,,190.0000,,,,,,,,,,,
french-cuff-cotton-twill-oxford,,,,,,,,,,,,,,,,msj000xs,,,,,,,,,,,,190.0000,,,,,,,,,,,
french-cuff-cotton-twill-oxford,,,,,,,,,,,,,,,,msj000xl,,,,,,,,,,,,190.0000,,,,,,,,,,,

```

Seems like it converts the price and sku correctly, but does not convert the weight. Any reason this might happen? Am I just confused?

JordanVerasamy:
Ugh I had added it and apparently commented it back out. Sorry fam. Works as expected.

lib/shopify_transporter/pipeline/magento/product/top_level_variant_attributes.rb
```diff
@@ -8,7 +8,10 @@ module Magento
       module Product
         class TopLevelVariantAttributes < Pipeline::Stage
           def convert(hash, record)
-            accumulator = TopLevelVariantAttributesAccumulator.new(record['variants'])
+            simple_product_in_magento_format = record['variants'].select do |product|
+              product['product_id'] == hash['product_id']
+            end[0]
```
fahia: nit: prefer `end.first` vs `end[0]`

spec/shopify_transporter/pipeline/magento/product/top_level_variant_attributes_spec.rb
```diff
@@ -9,14 +9,13 @@ module ShopifyTransporter::Pipeline::Magento::Product
...
-            }.deep_stringify_keys
-          ]
-        expect(expected_variants_in_shopify_format).to eq(variants_in_shopify_format)
+        expected_variants_in_shopify_format = {
+          sku: child_product['sku'],
+          grams: child_product['weight'],
+          price: child_product['price']
+        }.deep_stringify_keys
+
+        expect(variants_in_shopify_format).to include(expected_variants_in_shopify_format)
```
fahia: Would prefer the stringify keys to be on this line vs above as we usually do in other tests (it's been generally helpful too, IMO).

PR #48, opened 2018-09-26
Extract and convert product tags

zliu1020:
### What it does and why:
Extract and convert product tags

### Checklist:

### Demo:
screen shot 2018-09-26 at 11 12 12 am

---

### Related issues: https://github.com/Shopify/transporter_app/issues/1321
Closes:

### DIFF
--- a/lib/shopify_transporter/pipeline/magento/product/top_level_attributes.rb
+++ b/lib/shopify_transporter/pipeline/magento/product/top_level_attributes.rb
@@ -31,6 +31,7 @@ module ShopifyTransporter
               attributes['published_scope'] = published?(input) ? 'global' : ''
               attributes['published_at'] = published?(input) ? input['updated_at'] : ''
               attributes['images'] = append_images(input) if input['images'].present?
+              attributes['tags'] = product_tags(input) if input['tags'].present?
               attributes
             end
 
@@ -53,6 +54,12 @@ module ShopifyTransporter
               if label.is_a? String
                 label
               end
+            end
+
+            def product_tags(input)
+              input['tags'].each_with_object([]) do |tag, array|
+                array << tag['name']
+              end.join(', ')
             end
           end
         end
--- a/lib/shopify_transporter/pipeline/magento/product/top_level_attributes.rb
+++ b/lib/shopify_transporter/pipeline/magento/product/top_level_attributes.rb
@@ -57,9 +57,7 @@ module ShopifyTransporter
             end
 
             def product_tags(input)
-              input['tags'].each_with_object([]) do |tag, array|
-                array << tag['name']
-              end.join(', ')
+              input['tags'].map { |tag| tag['name'] }.join(', ')
             end
           end
         end
--- a/lib/shopify_transporter/pipeline/magento/product/top_level_variant_attributes.rb
+++ b/lib/shopify_transporter/pipeline/magento/product/top_level_variant_attributes.rb
@@ -8,12 +8,16 @@ module ShopifyTransporter
       module Product
         class TopLevelVariantAttributes < Pipeline::Stage
           def convert(hash, record)
-            return if record['variants'].nil?
+            return unless input_applies?(hash)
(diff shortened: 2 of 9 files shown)


lib/shopify_transporter/pipeline/magento/product/top_level_attributes.rb
```diff
@@ -54,6 +55,12 @@ def image_alt_text(label)
                 label
               end
             end
+
+            def product_tags(input)
+              input['tags'].each_with_object([]) do |tag, array|
```
JordanVerasamy: Is there a reason we can't just do `input['tags'].map { |tag| tag['name'] }.join(', ')`?

lib/shopify_transporter/pipeline/magento/product/top_level_variant_attributes.rb
```diff
@@ -8,6 +8,7 @@ module Magento
       module Product
         class TopLevelVariantAttributes < Pipeline::Stage
           def convert(hash, record)
+            return if record['variants'].nil?
```
JordanVerasamy: why is this needed now but wasn't before?

spec/shopify_transporter/exporters/magento/product_exporter_spec.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 require 'shopify_transporter/pipeline/stage'
-
+require 'pry'
```
JordanVerasamy: remove

JordanVerasamy:
i will also mention that i 🎩 ed this and it works great!

lib/shopify_transporter/pipeline/magento/product/top_level_attributes.rb
JordanVerasamy: you also could do something fancy with inject/reduce, but doing a map then a join is prob the most readable approach

zliu1020: I used ```each_with_object``` for no particular reason, it was just on the top of my head and ```map``` didn't come to my mind. You're absolutely right and ```map``` seems to be more readable than ```each_with_object```

lib/shopify_transporter/pipeline/magento/product/top_level_variant_attributes.rb
fahia: Probably fails if anything called `simple_product_in_magento_format` as that assumes the presence of `variants`.

lib/shopify_transporter/pipeline/magento/product/top_level_attributes.rb
surajreddy: 👍 to Jordan's recommendation.

lib/shopify_transporter/pipeline/magento/product/top_level_variant_attributes.rb
surajreddy: I think we should remove this return: `return if record['variants'].nil?`

If the variants array doesn't exist, then I suggest that we should set it to an empty array in the Product Record Builder/handle it in the Record Builder. We can also handle it in this file potentially, but I suggest tackling this in a different PR.

For this PR, I suggest removing this line.

zliu1020: This line of code should've been added when I was working on the ticket about top level variant fields. Converting top level variants makes no sense if there're no variants in the first place. I added this line just for safety.

zliu1020:
I've made changes so that the check against if the input product has a parent is the first thing to be carried out in top level variant conversion. Previously this check won't happen until the accumulation action is about to start. Also as pointed out by @surajreddy , the ProductRecordBuilder class now initializes the parent record with an empty variant array. In this way the downstream code which uses the ProductRecordBuilder doesn't need to bother checking against the existence of variants array.

PR #50, opened 2018-09-26
Extract Product Options from Magento

surajreddy:
### What it does and why:
- Extracts variant options for configurable magento products in shopify's format.
- Adds database exporter and database cache classes to assist with working with the magento databases.
- Uses a `/cache/` folder for the database cache. This folder can be deleted to fetch tables again.

Protip: This is a large PR. I suggest using the commits to see the PR built incrementally.

### Checklist:


lib/shopify_transporter/exporters/magento/product_exporter.rb
```diff
@@ -62,20 +63,45 @@ def merge_simple_product_with_parent(product, product_mappings)
...
+          @product_options.shopify_option_names(product[:product_id])
+        end
+
+        def info_for(product)
+          additional_attributes = if product[:parent_id]
+            @product_options.option_names_for_soap(product[:parent_id])
+          end
+
+          attributes = if additional_attributes
+            { 'additional_attributes' => { item: additional_attributes } }
```
fahia: Any reason why this is nested?

spec/shopify_transporter/exporters/magento/product_exporter_spec.rb
```diff
@@ -366,23 +395,342 @@ module Magento
...
+                  parent_id: '801',
+                  images: [],
+                  tags: [],
+                  another_attribute: 'another_value',
+                  inventory_quantity: 5,
+                  option1_name: 'Color',
+                  option1_value: 'White',
+                  option2_name: 'Size',
+                  option2_value: 'XS',
+                  additional_attributes: {
```
fahia: any reason why this can't be something like:
(name could be better but you get the idea)
```yaml
option_details: [
    {},
    {}
]
```
instead of `additional_attributes -> item ->` ?

lib/shopify_transporter/exporters/magento/product_options.rb
```diff
@@ -0,0 +1,97 @@
...
+          export_required_tables
+        end
+
+        def shopify_option_names(parent_product_id)
+          option_names(parent_product_id).each_with_index.with_object({}) do |(option_name, index), obj|
+            obj["option#{index + 1}_name".to_sym] = option_name
+          end
+        end
+
+        def option_names_for_soap(parent_product_id)
```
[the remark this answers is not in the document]
surajreddy: The `for_soap` indicates that it's specially formatted for soap (we need them to be `lowercase` for soap). Without this it doens't work right.

Should I just rename to `lowercase_option_names`?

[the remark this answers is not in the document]
fahia: lowercase_option_names sounds reasonable 👍

lib/shopify_transporter/exporters/magento/product_exporter.rb
surajreddy: I thought without the block the variable will be undefined. I wanted it to be nil if the condition failed.

But I was wrong, I'll change it.

```
irb(main):001:0> a = 5 if false
=> nil
irb(main):002:0> a
=> nil
```

spec/shopify_transporter/exporters/magento/product_exporter_spec.rb
surajreddy: The `additional_attributes` is just the raw soap result we get from the product info soap call, the keys we care about are the `option1_name`, `option1_value`, etc.

We can remove the additional_attributes, but I left it in in case we find some other use for it in the conversion steps.

fahia: It looks like it's just adding extra nesting that makes the conversion step more complex. I'd just pull it up a few levels for ease of access and simplicity.

lib/shopify_transporter/exporters/magento/database_table_exporter.rb
```diff
@@ -0,0 +1,72 @@
...
+        def write_batch(export_file_path, headers, batch)
+          File.open(export_file_path, 'a') do |file|
+            batch.each do |row|
+              data = headers.map { |header| row[header] }
+              file << data.to_csv
+            end
+          end
+        end
+
+        def create_cache_dir_if_needed
```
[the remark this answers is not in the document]
surajreddy: 👍 great suggestions! Thanks!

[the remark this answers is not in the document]
JordanVerasamy: :+1: to fahia's suggestions

lib/shopify_transporter/exporters/magento/product_exporter.rb
```diff
@@ -1,24 +1,27 @@
...
+          @database_table_exporter = DatabaseTableExporter.new(database_adapter)
+          @database_cache = DatabaseCache.new
+          @product_options = ProductOptions.new(@database_table_exporter, @database_cache)
         end
 
         def export
           $stderr.puts 'Starting export...'
-          products = base_products.map do |product|
+          products = apply_mappings(base_products)
+          products.map do |product|
```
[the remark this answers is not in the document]
surajreddy: 👍 that's fine as well - I tried separating it because it might be easier to read/follow the code if it was separated.

lib/shopify_transporter/exporters/magento/database_cache.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+require 'csv'
+require 'sequel'
+
+module ShopifyTransporter
+  module Exporters
+    module Magento
+      class DatabaseCache
```
fahia: TBH, I don't think this is the right time for this abstraction. Especially considering it's use. I don't think it's blocking but I just wanted to point that out. I could see this being just a part of the options class. It's basically only used within there making it pretty much isolated in its use.

JordanVerasamy: I vaguely thought this too. I just didn't feel it strongly enough to block this PR over it.

fahia: ^ non-blocking

surajreddy: 👍 Thanks for the feedback folks! 

This came before `ProductOptions` and I didn't think about it.

Personally I was thinking we should wrap `DatabaseCache` and `DatabaseExporter` into a `VirtualDatabase`/some middle group, and use that instead of 2 classes for the database. I didn't get around to it :(

JordanVerasamy: Like I said above -- I'm ok with merging things the way they evolved, and changing them in the future as needs come up or as we have better ideas. I wouldn't push hard for you to remove this abstraction.

PR #51, opened 2018-09-27
Extract and convert variant images

zliu1020:
### What it does and why:
- Code extracting variant images already exists
- Add a new pipeline stage to convert variant images

### Checklist:

### Demo:
screen shot 2018-09-27 at 11 26 07 am

---

### Related issues: https://github.com/Shopify/transporter_app/issues/1325
Closes:

### DIFF
--- a/lib/shopify_transporter/pipeline/magento/product/top_level_attributes.rb
+++ b/lib/shopify_transporter/pipeline/magento/product/top_level_attributes.rb
@@ -30,8 +30,12 @@ module ShopifyTransporter
               attributes['published'] = published?(input)
               attributes['published_scope'] = published?(input) ? 'global' : ''
               attributes['published_at'] = published?(input) ? input['updated_at'] : ''
+<<<<<<< HEAD
               append_images_to_current_record!(input) if input['images'].present?
+=======
+>>>>>>> extract and convert variant images
               attributes['tags'] = product_tags(input) if input['tags'].present?
+              append_images(input) if input['images'].present?
               attributes
             end
 
@@ -50,7 +54,11 @@ module ShopifyTransporter
               images.sort_by { |image| image['position'] }
             end
 
+<<<<<<< HEAD
             def append_images_to_current_record!(input)
+=======
+            def append_images(input)
+>>>>>>> extract and convert variant images
               images = construct_images(input)
               if @output['images'].present?
                 @output['images'].concat(images)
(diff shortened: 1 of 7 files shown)


lib/shopify_transporter/pipeline/magento/product/top_level_attributes.rb
```diff
@@ -50,6 +50,15 @@ def append_images(input)
               images.sort_by { |image| image['position'] }
             end
 
+            def append_images(input)
```
JordanVerasamy: Is there a way to do this without using mutation? What motivated the use of mutation here?

lib/shopify_transporter/pipeline/magento/product/variant_image.rb
```diff
@@ -0,0 +1,47 @@
...
+
+module ShopifyTransporter
+  module Pipeline
+    module Magento
+      module Product
+        class VariantImage < Pipeline::Stage
+          def convert(hash, record)
+            return unless input_applied?(hash)
+            accumulate(hash, record)
+            add_variant_img_to_parent_image_array(record, hash)
```
JordanVerasamy: The order of parameters should be consistent -- you have `hash, record` for the `convert` method, and then `record, hash` for the `add_variant_img_to_parent_image_array` method.

spec/shopify_transporter/pipeline/magento/product/variant_image_spec.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+require 'shopify_transporter/pipeline/magento/product/variant_image'
+
+module ShopifyTransporter::Pipeline::Magento::Product
+  RSpec.describe VariantImage, type: :helper do
```
JordanVerasamy: Add a test for when there are multiple images? I believe that the code path that concatenates images onto an existing image array is not being tested.

lib/shopify_transporter/pipeline/magento/product/top_level_attributes.rb
JordanVerasamy: At the _very_ least, I'd want it to be `append_images_to_output!` or something -- a more descriptive name that describes what it is doing and indicates its mutative nature.

spec/shopify_transporter/pipeline/magento/product/variant_image_spec.rb
fahia: image

We need to make sure this file is 100% covered

fahia:
re: img vs image. I don't care as long as it's consistent everywhere.

JordanVerasamy:
Nothing to add -- just a general +1 to Fahia's suggestions.

zliu1020:
@JordanVerasamy I've discussed with Fahia offline and want to keep you posted as well. The reason why I use the mutation in ```top_level_attributes.rb``` is because there is no way of making sure the parent product gets processed before its children does. Meanwhile, we cannot assure that ```TopLevelAttributes``` stage always come before other pipeline stages. If the child products are processed before the parent, we'd like to merge their image array with the parent images when it's the turn to convert parent images. This explains the ```@output['images'].concat(images)``` part. Let me know if this does not make sense to you.

PR #52, opened 2018-09-27
Write data incrementally to stdout

JordanVerasamy:
### What it does and why:

Print the output as we get it instead of all at the end. This means that if something goes wrong midway, instead of losing all the work done so far, the user can see what was done. This is a step in the right direction so that using the exporters for large data sets will be less painful.

### Checklist:

---

### Related issues:
Closes:
This is the first half of work related to https://github.com/Shopify/transporter_app/issues/1282

### DIFF
--- a/lib/shopify_transporter/exporters/exporter.rb
+++ b/lib/shopify_transporter/exporters/exporter.rb
@@ -18,19 +18,35 @@ module ShopifyTransporter
       end
 
       def run
-        data = Magento::MagentoExporter
-          .for(type: object_type)
-          .new(
-            soap_client: soap_client,
-            database_adapter: database_adapter
-          )
-          .export
-        puts JSON.pretty_generate(data) + $INPUT_RECORD_SEPARATOR
+        print_exported_objects
       end
 
       private
 
       attr_reader :config, :object_type
+
+      def print_exported_objects
+        $stderr.puts 'Starting export...'
+        puts '['
+        first = true
+        object_exporter.export do |object|
+          puts ',' unless first
+          first = false
+          $stderr.puts "Fetching #{object_type}: #{object[object_exporter.key]}..."
+          print '  ' + JSON.pretty_generate(object, object_nl: "\n  ")
+        end
+      ensure
+        print "\n]"
+      end
+
+      def object_exporter
+        @exporter ||= Magento::MagentoExporter
+          .for(type: object_type)
+          .new(
+            soap_client: soap_client,
+            database_adapter: database_adapter
+          )
+      end
 
       def soap_client
         Magento::Soap.new(
--- a/lib/shopify_transporter/exporters/magento/customer_exporter.rb
+++ b/lib/shopify_transporter/exporters/magento/customer_exporter.rb
@@ -9,15 +9,21 @@ module ShopifyTransporter
           @database_adapter = database_adapter
         end
 
+        def key
+          :customer_id
+        end
+
         def export
-          $stderr.puts 'Starting export...'
-          base_customers.map do |customer|
-            $stderr.puts "Fetching customer: #{customer[:customer_id]}..."
-            customer.merge(address_list: customer_address_list(customer[:customer_id]))
-          end.compact
+          base_customers.each do |customer|
+            yield with_attributes(customer)
+          end
         end
 
         private
+
+        def with_attributes(base_customer)
+          base_customer.merge(address_list: customer_address_list(base_customer[:customer_id]))
+        end
 
         def base_customers
           result = @client.call(:customer_customer_list, filters: nil).body
--- a/lib/shopify_transporter/exporters/magento/order_exporter.rb
+++ b/lib/shopify_transporter/exporters/magento/order_exporter.rb
@@ -9,15 +9,21 @@ module ShopifyTransporter
           @database_adapter = database_adapter
         end
 
+        def key
+          :increment_id
+        end
+
         def export
-          $stderr.puts 'Starting export...'
-          base_orders.map do |order|
-            $stderr.puts "Fetching order: #{order[:increment_id]}..."
-            order.merge(items: info_for(order[:increment_id]))
-          end.compact
+          base_orders.each do |order|
+            yield with_attributes(order)
+          end
         end
 
         private
+
+        def with_attributes(base_order)
+          base_order.merge(items: info_for(base_order[:increment_id]))
+        end
 
         def base_orders
           result = @client.call(:sales_order_list, filters: nil).body
--- a/lib/shopify_transporter/exporters/magento/product_exporter.rb
+++ b/lib/shopify_transporter/exporters/magento/product_exporter.rb
@@ -12,13 +12,14 @@ module ShopifyTransporter
           @intermediate_file_name = 'magento_product_mappings.csv'
         end
 
+        def key
+          :product_id
+        end
+
         def export
-          $stderr.puts 'Starting export...'
-          products = base_products.map do |product|
-            $stderr.puts "Fetching product: #{product[:product_id]}"
-            with_attributes(product)
-          end.compact
-          apply_mappings(products)
+          apply_mappings(base_products).each do |product|
+            yield with_attributes(product)
+          end
         end
 
         private
--- a/spec/shopify_transporter/exporters/magento/product_exporter_spec.rb
+++ b/spec/shopify_transporter/exporters/magento/product_exporter_spec.rb
@@ -1,6 +1,5 @@
 # frozen_string_literal: true
 require 'shopify_transporter/pipeline/stage'
-require 'pry'
 
 module ShopifyTransporter
   module Exporters
--- a/spec/shopify_transporter/exporters/magento/product_exporter_spec.rb
+++ b/spec/shopify_transporter/exporters/magento/product_exporter_spec.rb
@@ -1,10 +1,17 @@
 # frozen_string_literal: true
 require 'shopify_transporter/pipeline/stage'
+require 'pry'
 
 module ShopifyTransporter
   module Exporters
     module Magento
       RSpec.describe ProductExporter do
+        context '#key' do
+          it 'returns :product_id' do
+            expect(described_class.new.key).to eq(:product_id)
+          end
+        end
+
         context '#run'
           it 'retrieves configurable products from Magento using the SOAP API and returns the results' do
             soap_client = double("soap client")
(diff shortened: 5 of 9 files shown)


lib/shopify_transporter/exporters/exporter.rb
```diff
@@ -18,15 +18,14 @@ def initialize(config_filename, object_type)
...
       def run
-        data = Magento::MagentoExporter
+        Magento::MagentoExporter
           .for(type: object_type)
           .new(
             store_id: config['export_configuration']['store_id'],
             soap_client: soap_client,
             database_adapter: database_adapter
           )
           .export
```
JordanVerasamy: Maybe there's a better name for these functions now? They aren't just exporting anymore -- they're printing data to standard out.

lib/shopify_transporter/exporters/magento/customer_exporter.rb
```diff
@@ -12,14 +12,21 @@ def initialize(store_id: nil, soap_client: nil, database_adapter: nil)
 
         def export
```
JordanVerasamy: All three `SomethingExporter#export` functions are nearly the same, so there's potential for some OOP stuff. I'm leaving it simple for now and we can do some refactoring and stuff when the dust settles.

spec/spec_helper.rb
```diff
@@ -44,6 +44,15 @@ def capture(stream)
   end
   alias :silence :capture
 
+  def with_captured_stdout
```
fahia: The name for this smells like it's duplicate functionality to the capture method above.. Anyway to make this work?

lib/shopify_transporter/exporters/magento/customer_exporter.rb
```diff
@@ -12,14 +12,21 @@ def initialize(store_id: nil, soap_client: nil, database_adapter: nil)
...
           $stderr.puts 'Starting export...'
-          base_customers.map do |customer|
-            $stderr.puts "Fetching customer: #{customer[:customer_id]}..."
-            customer.merge(address_list: customer_address_list(customer[:customer_id]))
+          puts '['
```
fahia: Man, this entire thing is such a hack. I don't know of an alternative way to do this, though.

lib/shopify_transporter/exporters/magento/customer_exporter.rb
```diff
@@ -12,14 +12,21 @@ def initialize(store_id: nil, soap_client: nil, database_adapter: nil)
...
+          base_customers.each_with_index do |base_customer, i|
+            puts ',' unless i == 0
```
fahia: We can avoid the magic number by just saying `unless base_customer == base_customer.first`

lib/shopify_transporter/exporters/magento/customer_exporter.rb
```diff
@@ -12,14 +12,21 @@ def initialize(store_id: nil, soap_client: nil, database_adapter: nil)
...
+            $stderr.puts "Fetching customer: #{base_customer[:customer_id]}..."
+            print '  ' + JSON.pretty_generate(with_attributes(base_customer), object_nl: "\n  ")
```
fahia: Ouch.

spec/shopify_transporter/exporters/magento/product_exporter_spec.rb
```diff
@@ -345,22 +349,22 @@ module Magento
 
               expected_result = [
                 {
```
fahia: are all of these changes a result of the change to how we log to stdout?

spec/spec_helper.rb
JordanVerasamy: Shit. I didn't even notice that the above function exists. This can be removed safely. Good catch.

lib/shopify_transporter/exporters/magento/customer_exporter.rb
JordanVerasamy: Preaching to the choir. I'd love any alternative way to do this. At all.

It hurt me to write this code.

lib/shopify_transporter/exporters/magento/customer_exporter.rb
JordanVerasamy: I know... but I don't think there's a better way to generate output this pretty.

spec/shopify_transporter/exporters/magento/product_exporter_spec.rb
JordanVerasamy: Yes. Everything needs to be stringified all the way down because of the fact that we now print to stdout and then use `JSON.parse` on the data that we grab from stdout. Unfortunately, `deep_stringify_keys` only works on `Hash`. The alternative is to write our own `stringify_data` or something that works on both `Hash` and `Array` even when they're nested in each other. I don't mind doing that if you prefer it.

fahia: No no this is fine. Just a question. I wouldn't want you to go embark on that just to keep this from changing. This still does its job.

lib/shopify_transporter/exporters/magento/customer_exporter.rb
fahia: I think a next step would be subsequent refactors where we strip out the duplicate code between each of these Exporter#export methods. We would then have a service object that does this printing to stdout and we can just feed it blocks to print or something. I'll play around with it, I think it would be a nice way to just ensure we've taken all the...pragmatic stuff and encapsulated somewhere.

JordanVerasamy: Sounds like a good plan.

lib/shopify_transporter/exporters/magento/customer_exporter.rb
```diff
@@ -12,14 +12,21 @@ def initialize(store_id: nil, soap_client: nil, database_adapter: nil)
...
-          base_customers.map do |customer|
-            $stderr.puts "Fetching customer: #{customer[:customer_id]}..."
-            customer.merge(address_list: customer_address_list(customer[:customer_id]))
+          puts '['
+          base_customers.each do |base_customer|
+            puts ',' unless base_customer == base_customers.first
+            $stderr.puts "Fetching customer: #{base_customer[:customer_id]}..."
+            print '  ' + JSON.pretty_generate(with_attributes(base_customer), object_nl: "\n  ")
           end.compact
+          print "\n]"
```
surajreddy: Add an `ensure` above this last print - in case there is an error or exception, within the `base_customers` loop, the file will still be complete.

lib/shopify_transporter/exporters/magento/customer_exporter.rb
```diff
@@ -12,14 +12,21 @@ def initialize(store_id: nil, soap_client: nil, database_adapter: nil)
...
         def export
           $stderr.puts 'Starting export...'
-          base_customers.map do |customer|
-            $stderr.puts "Fetching customer: #{customer[:customer_id]}..."
-            customer.merge(address_list: customer_address_list(customer[:customer_id]))
+          puts '['
+          base_customers.each do |base_customer|
+            puts ',' unless base_customer == base_customers.first
+            $stderr.puts "Fetching customer: #{base_customer[:customer_id]}..."
+            print '  ' + JSON.pretty_generate(with_attributes(base_customer), object_nl: "\n  ")
```
surajreddy: change `\n` to `#{$INPUT_RECORD_SEPERATOR}` or `#{$RS}`. `\n` is machine specific.

lib/shopify_transporter/exporters/magento/customer_exporter.rb
JordanVerasamy: Aha! Maybe I should be rescuing inside every iteration of the loop too. Good thinking.

lib/shopify_transporter/exporters/magento/order_exporter.rb
```diff
@@ -12,14 +12,21 @@ def initialize(store_id: nil, soap_client: nil, database_adapter: nil)
...
-          base_orders.map do |order|
-            $stderr.puts "Fetching order: #{order[:increment_id]}..."
-            order.merge(items: info_for(order[:increment_id]))
+          puts '['
+          base_orders.each do |base_order|
+            puts ',' unless base_order == base_orders.first
+            $stderr.puts "Fetching order: #{base_order[:increment_id]}..."
+            print '  ' + JSON.pretty_generate(with_attributes(base_order), object_nl: "\n  ")
           end.compact
+          print "\n]"
```
surajreddy: add `ensure` block.

lib/shopify_transporter/exporters/magento/product_exporter.rb
```diff
@@ -15,11 +15,14 @@ def initialize(store_id: nil, soap_client: nil, database_adapter: nil)
...
-            with_attributes(product)
+          puts '['
+          mapped_products = apply_mappings(base_products)
+          mapped_products.each do |mapped_product|
+            puts ',' unless mapped_product == mapped_products.first
+            $stderr.puts "Fetching product: #{mapped_product[:product_id]}..."
+            print '  ' + JSON.pretty_generate(with_attributes(mapped_product), object_nl: "\n  ")
           end.compact
-          apply_mappings(products)
+          print "\n]"
```
surajreddy: add `ensure` block.

lib/shopify_transporter/exporters/magento/order_exporter.rb
```diff
@@ -12,14 +12,21 @@ def initialize(store_id: nil, soap_client: nil, database_adapter: nil)
 
         def export
           $stderr.puts 'Starting export...'
-          base_orders.map do |order|
-            $stderr.puts "Fetching order: #{order[:increment_id]}..."
-            order.merge(items: info_for(order[:increment_id]))
+          puts '['
```
surajreddy: I think we should add tests that check the output works with multiple objects and not just one.

You mentioned this, but I'm hugely in favor of moving some of this code up out of each of the `export` methods.

What if we add a `print_exported_objects` method in `lib/shopify_transporter/exporters/exporter.rb` and made these export methods yield objects instead?

```
def run
  print_exported_objects
end

def print_exported_objects
  puts '['
  first_object = true
  Magento::MagentoExporter
    .for(type: object_type)
    .new(
      store_id: config['export_configuration']['store_id'],
      soap_client: soap_client,
      database_adapter: database_adapter
    )
    .export_objects do |object|
      if first_object
        puts ','
        first_object = false
      end
      print '  ' + JSON.pretty_generate(object, object_nl: "\n  ")
    end
  ensure
  puts ']'
end


# In exporters - we can probably use a better name.
def export_objects
  base_orders.each do |base_order|
    puts ',' unless base_order == base_orders.first
    $stderr.puts "Fetching order: #{base_order[:increment_id]}..."
    yield with_attributes(base_order)
  end
end
```

Then all the JSON code can be tested in one place.

lib/shopify_transporter/exporters/magento/customer_exporter.rb
surajreddy: remove the `compact` after `end`

lib/shopify_transporter/exporters/magento/order_exporter.rb
fahia: I like this idea. I'd go as far as to extract the logger to its own class. It's pretty standalone.

spec/shopify_transporter/exporters/magento/product_exporter_spec.rb
```diff
@@ -1,10 +1,17 @@
 # frozen_string_literal: true
 require 'shopify_transporter/pipeline/stage'
+require 'pry'
```
surajreddy: ☠️

PR #57, opened 2018-09-28
Add batch parameters and export data in batches

surajreddy:
### What it does and why:
- Retrieves data in batches to reduce load on the server, and to incrementally export data.

### Checklist:

### DIFF
--- a/exe/shopify_transporter
+++ b/exe/shopify_transporter
@@ -8,6 +8,7 @@ require 'thor/group'
 require 'shopify_transporter'
 require 'yaml'
 require 'io/console'
+require 'active_support/core_ext/string'
 
 module ShopifyTransporter
   class CLI < Thor
@@ -35,10 +36,19 @@ module ShopifyTransporter
     method_option :object, type: :string, required: true,
       desc: 'The object type (customer, product, or order) to extract.',
       enum: %w(customer product order)
+    method_option :first_id, type: :numeric, required: true,
+      desc: 'The first product_id, customer_id, or order_id for the range of objects you are trying to export.'
+    method_option :last_id, type: :numeric, required: true,
+      desc: 'The last product_id, customer_id, or order_id (inclusive) for the range of objects you are trying' \
+        ' to export.'
+    method_option :batch_size, type: :numeric, default: 1000,
+      desc: 'The number of records that are retrieved from the source platform at one time. ' \
+        'Use a smaller batch size if you are running into performance issues extracting your data.'
 
     desc 'extract', 'Extracts objects from a third-party platform into a format compatible for conversion.'
     def extract
-      exporter = Exporters::Exporter.new(options[:config], options[:object])
+      batch_config = options.slice('first_id', 'last_id', 'batch_size')
+      exporter = Exporters::Exporter.new(options[:config], options[:object], batch_config)
       exporter.run
     rescue Exporters::ExportError => error
       extract_error(error.message)
--- a/lib/shopify_transporter/exporters/magento/customer_exporter.rb
+++ b/lib/shopify_transporter/exporters/magento/customer_exporter.rb
@@ -26,8 +26,13 @@ module ShopifyTransporter
         end
 
         def base_customers
-          result = @client.call(:customer_customer_list, filters: nil).body
-          result[:customer_customer_list_response][:store_view][:item] || []
+          Enumerator.new do |enumerator|
+            @client.call_in_batches(method: :customer_customer_list, batch_index_column: 'customer_id').each do |batch|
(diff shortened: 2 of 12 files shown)


exe/shopify_transporter
```diff
@@ -35,10 +35,17 @@ module ShopifyTransporter
     method_option :object, type: :string, required: true,
       desc: 'The object type (customer, product, or order) to export.',
       enum: %w(customer product order)
+    method_option :first_id, type: :numeric, required: true,
+      desc: 'The starting id of the object type you are trying to export.'
```
JordanVerasamy: These descriptions are not very descriptive. I'm not sure how to improve them, but I definitely think they need attention.

fahia: I'm not sure what's the best wording here. Might be worth getting into the range query args in the new wiki.

surajreddy: Based on some input from Jenna, I'll change this to:

```
first_id

The first product_id, customer_id, or order number for the range of objects you are trying to export.


last_id

The last product_id, customer_id, or order number for the range of objects you are trying to export.


batch_size

The number of records that are retrieved from the source platform at one time. Use a smaller batch size if you are running into performance issues extracting your data.
```

JordanVerasamy: Great updates.

JordanVerasamy:
The changes are good. Once this is tophatted and stuff I'll feel comfortable merging this.

lib/shopify_transporter/exporters/magento/customer_exporter.rb
```diff
@@ -26,8 +26,13 @@ def with_attributes(base_customer)
         end
 
         def base_customers
-          result = @client.call(:customer_customer_list, filters: nil).body
-          result[:customer_customer_list_response][:store_view][:item] || []
+          Enumerator.new do |enumerator|
+            @client.call_in_batches(method: :customer_customer_list, batch_index_column: 'customer_id').each do |batch|
+              result = batch.body[:customer_customer_list_response][:store_view][:item] || []
+              result = [result] unless result.class == Array
```
mimperatore: More idiomatic: `result is_a? Array`, or:

```
result = [*batch.body[:customer_customer_list_response][:store_view][:item]]
```

exe/shopify_transporter
```diff
@@ -35,10 +35,18 @@ module ShopifyTransporter
...
+    method_option :last_id, type: :numeric, required: true,
+      desc: 'The last product_id, customer_id, or order_id for the range of objects you are trying to export.'
+    method_option :batch_size, type: :numeric, default: 1000,
+      desc: 'The number of records that are retrieved from the source platform at one time. '\
+        'Use a smaller batch size if you are running into performance issues extracting your data.'
 
     desc 'export', 'Exports objects from a third-party platform into a format compatible for conversion.'
     def export
-      exporter = Exporters::Exporter.new(options[:config], options[:object])
+      batch_config = options.select { |(key, _)| %w(first_id last_id batch_size).include? key }
```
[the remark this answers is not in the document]
surajreddy: We don't have access to slice with vanilla Ruby, we will need to bring in `require 'active_support/core_ext/string'`

I think that's a good idea/I'll go ahead and do this.

lib/shopify_transporter/exporters/magento/soap.rb
```diff
@@ -7,21 +7,55 @@ module ShopifyTransporter
...
+        MAX_RETRIES = 4
+
+        def initialize(hostname: '', username: '', api_key: '', batch_config:)
           @hostname = hostname
           @username = username
           @api_key = api_key
+          @batch_config = batch_config
+        end
+
+        def call(method, params)
```
mimperatore: To be consistent with the interface of `call_in_batches`, should we make this also `method:, params:`?

mimperatore: Or conversely changing `call_in_batches` _not_ to use named parameters?  Most methods are not using names params, so perhaps that should be the convention we go with.

spec/shopify_transporter/exporters/magento/soap_spec.rb
```diff
@@ -0,0 +1,193 @@
...
+
+            attempt = 0
+            expect(mock_client).to receive(:call).with(
+              :test_call,
+              message: {
+                session_id: '123',
+              },
+            ) do |args|
+              attempt += 1
+              if attempt < 5
```
mimperatore: Missed this one, I think.

lib/shopify_transporter/exporters/magento/soap.rb
surajreddy: Done!

PR #58, opened 2018-09-28
display warnings if the product has more than 3 options

zliu1020:
### What it does and why:
For now, CLI will display warnings when the product has more than three options. Only the first 3 options will get converted.

### Checklist:

### Demo:
screen shot 2018-09-28 at 3 24 32 pm
screen shot 2018-09-28 at 3 24 40 pm

---

### Related issues:https://github.com/Shopify/transporter_app/issues/1278
Closes:

### DIFF
--- a/lib/shopify_transporter/pipeline/magento/product/top_level_attributes.rb
+++ b/lib/shopify_transporter/pipeline/magento/product/top_level_attributes.rb
@@ -7,19 +7,25 @@ module ShopifyTransporter
     module Magento
       module Product
         class TopLevelAttributes < Pipeline::Stage
-          MAX_NUM_OF_OPTIONS_ACCEPTED = 3
-
           def convert(hash, record)
-            if has_too_many_options?(hash)
-              $stderr.puts "Warning: Product #{hash['product_id']} has too many options."\
-              " Only the first #{MAX_NUM_OF_OPTIONS_ACCEPTED} options will be converted."
-            end
+            warn_if_too_many_options(hash)
             accumulator = TopLevelAttributesAccumulator.new(record)
             accumulator.accumulate(hash)
           end
 
-          def has_too_many_options?(input)
+          private
+
+          MAX_NUM_OF_OPTIONS_ACCEPTED = 3
+
+          def too_many_options?(input)
             input["option#{MAX_NUM_OF_OPTIONS_ACCEPTED + 1}_name"].present?
+          end
+
+          def warn_if_too_many_options(input)
+            if too_many_options?(input)
+              $stderr.puts "Warning: Product #{input['product_id']} has too many options."\
+              " Only the first #{MAX_NUM_OF_OPTIONS_ACCEPTED} options will be converted."
+            end
           end
 
           class TopLevelAttributesAccumulator < Shopify::AttributesAccumulator
--- a/lib/shopify_transporter/pipeline/magento/product/top_level_attributes.rb
+++ b/lib/shopify_transporter/pipeline/magento/product/top_level_attributes.rb
@@ -7,9 +7,19 @@ module ShopifyTransporter
     module Magento
       module Product
         class TopLevelAttributes < Pipeline::Stage
+          MAX_NUM_OF_OPTIONS_ACCEPTED = 3
+
           def convert(hash, record)
+            if has_too_many_options?(hash)
+              $stderr.puts "Warning: Product #{hash['product_id']} has too many options."\
+              " Only the first #{MAX_NUM_OF_OPTIONS_ACCEPTED} options will be converted."
+            end
             accumulator = TopLevelAttributesAccumulator.new(record)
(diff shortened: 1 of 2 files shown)


lib/shopify_transporter/pipeline/magento/product/top_level_attributes.rb
```diff
@@ -7,11 +7,21 @@ module Pipeline
     module Magento
       module Product
         class TopLevelAttributes < Pipeline::Stage
+          MAX_NUM_OF_OPTIONS_ACCEPTED = 3
```
JordanVerasamy: propose `MAX_OPTION_COUNT` as name. shorter and more descriptive IMO

lib/shopify_transporter/pipeline/magento/product/top_level_attributes.rb
```diff
@@ -7,11 +7,21 @@ module Pipeline
...
+
           def convert(hash, record)
+            if has_too_many_options?(hash)
```
JordanVerasamy: suggest `too_many_options?` since i dont think `has` is offering any extra info here

spec/shopify_transporter/pipeline/magento/product/top_level_attributes_spec.rb
```diff
@@ -92,6 +92,29 @@ module ShopifyTransporter::Pipeline::Magento::Product
           expect(shopify_product.deep_stringify_keys).to include(expected_option_data.deep_stringify_keys)
         end
 
+        it 'ignore the rest of options when there are more than three options' do
```
JordanVerasamy: should be `ignores`, not `ignore`

spec/shopify_transporter/pipeline/magento/product/top_level_attributes_spec.rb
```diff
@@ -92,6 +92,29 @@ module ShopifyTransporter::Pipeline::Magento::Product
...
+              },
+              {
+                'name': 'option2',
+              },
+              {
+                'name': 'option3',
+              },
+            ],
+          }
+          expect(shopify_product.deep_stringify_keys).to include(expected_option_data.deep_stringify_keys)
```
JordanVerasamy: because you're using `include`, you aren't actually testing that the other options got ignored. i suggest using an exact equality check for this test, or using `not_to include` on the fourth option.

JordanVerasamy: https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/include-matcher

lib/shopify_transporter/pipeline/magento/product/top_level_attributes.rb
```diff
@@ -7,11 +7,21 @@ module Pipeline
...
+              $stderr.puts "Warning: Product #{hash['product_id']} has too many options."\
+              " Only the first #{MAX_NUM_OF_OPTIONS_ACCEPTED} options will be converted."
+            end
```
surajreddy: I suggest moving the whole if block to a private method as well. Maybe `warn_if_too_many_options`?

lib/shopify_transporter/pipeline/magento/product/top_level_attributes.rb
```diff
@@ -8,10 +8,26 @@ module Magento
       module Product
         class TopLevelAttributes < Pipeline::Stage
           def convert(hash, record)
+            warn_if_too_many_options(hash)
```
fahia: this would look more ruby if we did something like:
```ruby
warn(hash) if too_many_options?(hash)
```
Sucks that both `warn` and `too_many_options?` require the hash param. Might be nice to make it an instance method so that we could do `warn if too_many_options?`

lib/shopify_transporter/pipeline/magento/product/top_level_attributes.rb
```diff
@@ -8,10 +8,26 @@ module Magento
...
             accumulator = TopLevelAttributesAccumulator.new(record)
             accumulator.accumulate(hash)
           end
 
+          private
+
+          MAX_OPTION_COUNT = 3
```
surajreddy: Minor: I suggest moving `MAX_OPTION_COUNT` to the top above convert because it ends up being public. Or you can declare it private by doing this:
```ruby
MAX_OPTION_COUNT = 3
private_constant :MAX_OPTION_COUNT
```

lib/shopify_transporter/pipeline/magento/product/top_level_attributes.rb
surajreddy: We should also add this to `variant_attributes`

surajreddy: 🔥 this is a great suggestion.

PR #60, opened 2018-10-01
Use weight as the attribute for variant weight instead of grams

surajreddy:
### What it does and why:
- Uses `variant_weight` as the column for simple products' weight instead of grams.
    - This uses the default weight unit of the Shopify store as the weight unit when importing instead of grams because this allows more flexibility when importing weight.

### Checklist:


lib/shopify_transporter/shopify/product.rb
```diff
@@ -33,7 +33,8 @@ def header
             'Metafield Value Type', 'Variant Grams', 'Variant Inventory Tracker', 'Variant Inventory Qty',
```
zliu1020: `'Variant Grams'` should be removed if we're replacing it with `'Variant Weight'`

spec/files/transporter_csv_templates/products.csv
```diff
@@ -1,8 +1,8 @@
...
-hightop-sports-sneaker,,,,,,,,true,2007-12-31T19:00:00-05:00,,Black,,10,,,a-456,,,global,lace_style,normal,string,345,shopify,150,deny,manual,20.00,15.00,true,true,,,,,,https://cdn.shopify.com/shopify-marketing_assets/static/tobias-lutke-shopify.jpg,g,
-hightop-sports-sneaker,,,,,,,,true,2007-12-31T19:00:00-05:00,,Blue,,6,,,b-001,,,global,tread_type,rugged,string,345,shopify,150,deny,manual,30.00,15.00,true,true,,,,,,https://cdn.shopify.com/shopify-marketing_assets/static/tobias-lutke-shopify.jpg,g,
+Handle,Title,Body (HTML),Vendor,Type,Tags,Template Suffix,Published Scope,Published,Published At,Option1 Name,Option1 Value,Option2 Name,Option2 Value,Option3 Name,Option3 Value,Variant SKU,Metafields Global Title Tag,Metafields Global Description Tag,Metafield Namespace,Metafield Key,Metafield Value,Metafield Value Type,Variant Grams,Variant Inventory Tracker,Variant Inventory Qty,Variant Inventory Policy,Variant Fulfillment Service,Variant Price,Variant Compare At Price,Variant Requires Shipping,Variant Taxable,Variant Barcode,Image Attachment,Image Src,Image Position,Image Alt Text,Variant Image,Variant Weight,Variant Weight Unit,Variant Tax Code
```
zliu1020: Same thing here. We should get rid of `'Variant Grams'`

spec/shopify_transporter/shopify/product_spec.rb
```diff
@@ -29,7 +29,7 @@ module Shopify
             variant_grams variant_inventory_tracker variant_inventory_qty variant_inventory_policy
```
zliu1020: Need to get rid of `variant_grams`

zliu1020:
I've chatted with @surajreddy offline and he stated that the ticket is about adding support for `Variant Weight` instead of replacing `Variant Grams` with `Variant Weight`. In this case all code looks good to me!

PR #74, opened 2018-10-04
 return empty cell instead of empty string when published_at is not present

JordanVerasamy:
### What it does and why:

If we pass `""` as a published at date to the Transporter app, it fails. So we should just.. not do that.

### Checklist:

---

### Related issues:
Closes:

### DIFF
--- a/lib/shopify_transporter/pipeline/magento/product/top_level_attributes.rb
+++ b/lib/shopify_transporter/pipeline/magento/product/top_level_attributes.rb
@@ -49,7 +49,7 @@ module ShopifyTransporter
               append_images_to_current_record!(input) if input['images'].present?
               attributes['tags'] = product_tags(input) if input['tags'].present?
               attributes['options'] = product_options(input) if input['option1_name'].present?
-              attributes
+              attributes.compact
             end
 
             def published?(input)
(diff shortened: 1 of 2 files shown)


lib/shopify_transporter/pipeline/magento/product/top_level_attributes.rb
```diff
@@ -45,11 +45,11 @@ def attributes_from(input)
               attributes = map_from_key_to_val(COLUMN_MAPPING, input)
               attributes['published'] = published?(input)
               attributes['published_scope'] = published?(input) ? 'global' : ''
-              attributes['published_at'] = published?(input) ? input['updated_at'] : ''
+              attributes['published_at'] = published?(input) ? input['updated_at'] : nil
```
fahia: Why not just:
```ruby
attributes['published_at'] = input['updated_at'] if published?(input)

File: lib/shopify_transporter/pipeline/magento/product/top_level_attributes.rb [hunk 2]
```diff
@@ -45,11 +45,11 @@ def attributes_from(input)
...
               append_images_to_current_record!(input) if input['images'].present?
               attributes['tags'] = product_tags(input) if input['tags'].present?
               attributes['options'] = product_options(input) if input['option1_name'].present?
-              attributes
+              attributes.compact
```
fahia: can get rid of this with above change

lib/shopify_transporter/pipeline/magento/product/top_level_attributes.rb
JordanVerasamy: huh idk how i didnt think of that. thanks

Issue #75, opened 2018-10-04
Some top level attributes of orders are not converted in CSV file


zliu1020:
Columns that are not mapped yet:
 - [] Financial Status: Needs investigation - Not sure how to determine if payments are ``voided`` and ``authorized`` in Magento
 - [x] Fulfillment Status: Done
 - [x] Currency: Done
 - [x] Buyer Accepts Marketing: No equivalent in Magento 
 - [x] Cancel Reason: No equivalent in Magento 
 - [x] Cancelled At: Done
 - [x] Closed At: Done
 - [x] Tags: No equivalent in Magento 
 - [x] Note: No equivalent in Magento 
 - [x] Phone: Done
 - [x] Referring Site: No equivalent in Magento   

Some address attributes are also missing:
- [x] Shipping Company
- [x] Shipping Name
- [x] Billing Company
- [x] Billing Name

PR #76, opened 2018-10-05
Add more support for order top level attributes

zliu1020:
### What it does and why:

See https://github.com/Shopify/shopify_transporter/issues/75 for context

### Checklist:

---

### Related issues: https://github.com/Shopify/shopify_transporter/issues/75
Closes:

### DIFF
--- a/lib/shopify_transporter/exporters/magento/customer_exporter.rb
+++ b/lib/shopify_transporter/exporters/magento/customer_exporter.rb
@@ -15,10 +15,12 @@ module ShopifyTransporter
 
         def export
           base_customers.each do |customer|
-            yield with_attributes(customer)
-          rescue Savon::Error => e
-            print_customer_details_error(customer, e)
-            yield customer
+            begin
+              yield with_attributes(customer)
+            rescue Savon::Error => e
+              print_customer_details_error(customer, e)
+              yield customer
+            end
           end
         end
 
--- a/lib/shopify_transporter/exporters/magento/customer_exporter.rb
+++ b/lib/shopify_transporter/exporters/magento/customer_exporter.rb
@@ -15,12 +15,10 @@ module ShopifyTransporter
 
         def export
           base_customers.each do |customer|
-            begin
-              yield with_attributes(customer)
-            rescue Savon::Error => e
-              print_customer_details_error(customer, e)
-              yield customer
-            end
+            yield with_attributes(customer)
+          rescue Savon::Error => e
+            print_customer_details_error(customer, e)
+            yield customer
           end
         end
 
--- a/lib/shopify_transporter/pipeline/magento/order/addresses_attribute.rb
+++ b/lib/shopify_transporter/pipeline/magento/order/addresses_attribute.rb
@@ -31,13 +31,21 @@ module ShopifyTransporter
             {
               first_name: address_attrs['firstname'],
               last_name: address_attrs['lastname'],
+              name: name(address_attrs),
               phone: address_attrs['telephone'],
               address1: address_attrs['street'],
               city: address_attrs['city'],
               province_code: address_attrs['region'],
               zip: address_attrs['postcode'],
               country_code: address_attrs['country_id'],
+              company: address_attrs['company'],
             }
+          end
+
+          def name(address_attrs)
+            if address_attrs['firstname'].present? && address_attrs['lastname'].present?
+              address_attrs['firstname'] + ' ' + address_attrs['lastname']
+            end
           end
         end
       end
--- a/lib/shopify_transporter/pipeline/magento/order/top_level_attributes.rb
+++ b/lib/shopify_transporter/pipeline/magento/order/top_level_attributes.rb
@@ -43,6 +43,53 @@ module ShopifyTransporter
             }.stringify_keys
           end
 
+          def financial_status(hash)
+            order_state = hash['state']
+            status = nil
+
+            if order_state == 'Pending Payment'
+              status = 'pending'
+            elsif paid?(hash)
+              status = 'paid'
+            elsif partially_paid?(hash)
+              status = 'partially_paid'
+            elsif partially_refunded?(hash)
+              status = 'partially_refunded'
+            elsif refunded?(hash)
+              status = 'refunded'
+            end
+            status
+          end
+
+          def fulfillment_status(hash)
+            total_qty_ordered = hash['total_qty_ordered'].to_i
+            total_qty_shipped = total_qty_shipped(hash)
+            status = nil
+
+            if total_qty_shipped == total_qty_ordered
+              status = 'fulfilled'
+            elsif total_qty_shipped > 0 && total_qty_shipped < total_qty_ordered
+              status = 'partial'
+            end
+            status
+          end
+
+          def cancelled_at(hash)
+            timestamp(hash, 'canceled') if cancelled?(hash)
+          end
+
+          def closed_at(hash)
+            timestamp(hash, 'closed') if closed?(hash)
+          end
+
+          def cancelled?(hash)
+            hash['state'] == 'canceled'
+          end
+
+          def closed?(hash)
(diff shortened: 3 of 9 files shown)


JordanVerasamy:
this code is looking sick btw

zliu1020:
### Issues: 
 - Support for ``financial_status`` is not complete. Not sure how to determine if payments are ``voided`` and ``authorized`` in Magento.
 - Setting ``closed_at`` or ``cancelled_at`` timestamps in CSV won't make them in status of closed/cancelled after being imported using transporter app. This is because ``closed_at`` and ``cancelled_at`` are read-only fields and Shopify API will set them to ``null``. One potential solution is to explicitly call Shopify API to cancel this order.

@JordanVerasamy @surajreddy @fahia Any thoughts on this?

JordanVerasamy:
regarding your issues: i do not think those belong in this PR. this PR is a notable improvement over the current behaviour, so we should make it work and merge it and worry about additional scope separately.

lib/shopify_transporter/pipeline/magento/order/top_level_attributes.rb
```diff
@@ -25,13 +34,79 @@ def convert(hash, record)
...
           end
+
+          def financial_status(hash)
+            total_price = hash['grand_total'].to_i
+            total_paid = hash['total_paid'].present? ? hash['total_paid'].to_i : 0
+            total_refunded = hash['total_refunded'].present? ? hash['total_refunded'].to_i : 0
+            order_state = hash['state']
+            status = nil
+
+            if order_state == 'Pending Payment'
```
fahia: wow

JordanVerasamy: yep i had a similar conversation with ziyan

the code is pretty much as good as it's possible to be. it's unfortunate.

fahia:
> regarding your issues: i do not think those belong in this PR. this PR is a notable improvement over the current behaviour, so we should make it work and merge it and worry about additional scope separately.

+1

lib/shopify_transporter/pipeline/magento/order/top_level_attributes.rb
JordanVerasamy: one tiny thing that might be an improvement (or might not, so give me your honest opinion):

change 

```
elsif total_price == total_paid && total_refunded == 0
              status = 'paid'
```

to

```
elsif paid?(hash)
              status = 'paid'
```

this allows you to move the logic from lines 48 to 51 into multiple small helper functions instead of at the top of this one, and allows you to make the big ugly section from lines 54 to 64 much more readable.

on the other hand, it means you're reading the contents of the hash in a place that is more separated from where that content matters, which is an extra layer of indirection that maybe isn't providing value. thoughts?

lib/shopify_transporter/pipeline/magento/order/addresses_attribute.rb
```diff
@@ -31,12 +31,14 @@ def address_attributes(input, prefix)
             {
               first_name: address_attrs['firstname'],
               last_name: address_attrs['lastname'],
+              name: address_attrs['firstname'] + ' ' + address_attrs['lastname'],
```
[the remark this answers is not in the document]
surajreddy: My reasoning is that `first_name` and `last_name` are more precise - if we have this information, then `name` is just an extra field. I don't think it's needed.

Does the order show up strangely in Shopify without `name` set?

lib/shopify_transporter/pipeline/magento/order/top_level_attributes.rb
fahia: +1 on this. For the purposes of readability, it might be better to actually wrap each of these conditions in the financial status name as a method: `paid?` `partially_paid?` etc. This will help with readability in the long run.

zliu1020: Thanks for the input! I'll extract these conditions into small methods.

lib/shopify_transporter/pipeline/magento/order/addresses_attribute.rb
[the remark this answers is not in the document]
zliu1020: I believe there is no difference if the order does not have ``name`` field. I'll get rid of this field.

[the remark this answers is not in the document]
surajreddy: Sounds good - I don't feel strongly about this if it doesn't affect how the data looks in Shopify.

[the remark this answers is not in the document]
zliu1020: LOL maybe I won't bother removing it then. Just too lazy to update relevant test cases.

lib/shopify_transporter/pipeline/magento/order/top_level_attributes.rb
```diff
@@ -25,13 +34,112 @@ def convert(hash, record)
...
+          end
+
+          def timestamp(hash, status)
+            return unless hash.dig('items', 'result', 'status_history', 'item').present?
+            timestamp = nil
+
+            history_items = hash['items']['result']['status_history']['item']
+            return history_items['created_at'] if history_items.is_a?(Hash) && history_items['status'] == status
+
+            history_items.each do |status_history|
```
surajreddy: Minor: Is the intent here to select the last `history_item` that matches the status?

How do you feel about selecting the last item and then setting this?

```
last_applicable_history_item = history_items.select { |history_item| history_item['status'] == status }.last
timestamp = last_applicable_history_item['created_at'] unless last_applicable_history_item.nil?
```

The benefit of this is just doing the assignment once, but because we don't expect there to be a lot of history items, I think the current code is fine/feel free to leave as is.

zliu1020: Good point! I've updated the code.

surajreddy:
I'm noticing an issue with total discounts - it is extracted correctly to the csv file, but it doesn't show up in the shopify order after being imported.

surajreddy:
Logged an issue for total_discounts, I suggest removing them from this PR: https://github.com/Shopify/shopify_transporter/issues/82

lib/shopify_transporter/pipeline/magento/order/top_level_attributes.rb
```diff
@@ -25,13 +34,111 @@ def convert(hash, record)
...
+          def total_refunded(hash)
+            return hash['total_refunded'].to_i if hash['total_refunded'].present?
+            0
+          end
+
+          def paid?(hash)
+            total_price(hash) == total_paid(hash) && total_refunded(hash) == 0
+          end
+
+          def partially_paid?(hash)
```
surajreddy: minor/this is a preference:
I suggest moving the method definitions for these methods to below the method where they are used. This makes it so that developers reading the code can ignore the lower level implementations unless they are interested in it.

zliu1020: Good point! I've relocated all helper functions to the bottom of the class.

lib/shopify_transporter/exporters/magento/customer_exporter.rb
```diff
@@ -15,12 +15,10 @@ def key
 
         def export
           base_customers.each do |customer|
-            begin
-              yield with_attributes(customer)
-            rescue Savon::Error => e
-              print_customer_details_error(customer, e)
-              yield customer
-            end
+            yield with_attributes(customer)
```
surajreddy: We ran into this with @fahia's pr as well - this rubocop suggestion is actually broken/is wrong because the code doesn't work anymore.

I suggest seeing if we can skip this check with @fahia.

zliu1020: I see. Thanks for pointing it out! I'll revert the previous commit.

Issue #68, opened 2018-10-03
When extracting data, skip individual objects that raise errors and continue with the remaining objects


surajreddy:
When extracting data, skip individual objects that raise errors and continue with the remaining objects.

We make additional soap calls for each object (order, customer, product). Sometimes, these additional calls fail. We should skip these objects and continue with the rest.

Issue #67, opened 2018-10-03
When extracting data, skip batches that raise errors


surajreddy:
When extracting data, skip batches that raise errors and continue with the remaining batches of data.

PR #77, opened 2018-10-09
Skip an object if subsequent api calls fail for that object

fahia:
### What it does and why:
- When building an object requires multiple requests, fail and move onto the next object if any of the calls happen to fail. Log this to stderr to let the user know.

### Checklist:

---

### Related issues:
Closes: https://github.com/Shopify/shopify_transporter/issues/68, https://github.com/Shopify/shopify_transporter/issues/67

### DIFF
--- a/lib/shopify_transporter/exporters/magento/customer_exporter.rb
+++ b/lib/shopify_transporter/exporters/magento/customer_exporter.rb
@@ -15,7 +15,12 @@ module ShopifyTransporter
 
         def export
           base_customers.each do |customer|
-            yield with_attributes(customer)
+            begin
+              yield with_attributes(customer)
+            rescue Savon::Error => e
+              print_customer_details_error(customer, e)
+              yield customer
+            end
           end
         end
 
@@ -37,6 +42,20 @@ module ShopifyTransporter
 
         def customer_address_list(customer_id)
           @client.call(:customer_address_list, customer_id: customer_id).body
+        end
+
+        def print_customer_details_error(customer, e)
+          $stderr.puts '***'
+          $stderr.puts 'Warning:'
+          $stderr.puts "Encountered an error with fetching details for customer with id: #{customer[:customer_id]}"
+          $stderr.puts JSON.pretty_generate(customer)
+          $stderr.puts 'The exact error was:'
+          $stderr.puts "#{e.class}: "
+          $stderr.puts e.message
+          $stderr.puts '-'
+          $stderr.puts "Exporting the customer (#{customer[:customer_id]}) without its details."
+          $stderr.puts 'Continuing with the next customer.'
+          $stderr.puts '***'
         end
       end
     end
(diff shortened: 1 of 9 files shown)


lib/shopify_transporter/exporters/magento/soap.rb
```diff
@@ -33,10 +33,14 @@ def call_in_batches(method:, params: {}, batch_index_column:)
...
-                method,
-                params.merge(batching_filter(current_id, end_of_range, batch_index_column)),
-              )
+              begin
+                enumerator << call(
+                  method,
+                  params.merge(batching_filter(current_id, end_of_range, batch_index_column)),
+                )
+              rescue Savon::Error
+                $stderr.puts "Skipping batch: #{current_id}..#{end_of_range} after #{MAX_RETRIES} retries because of an error."
```
surajreddy: Do you mind adding the error details here as well?
```
 $stderr.puts 'The exact error was:'
... e.class + e.message
```

lib/shopify_transporter/exporters/magento/customer_exporter.rb
```diff
@@ -38,6 +43,20 @@ def base_customers
         def customer_address_list(customer_id)
           @client.call(:customer_address_list, customer_id: customer_id).body
         end
+
+        def print_customer_details_error(customer, e)
```
surajreddy: Do you anticipate the error handling to change for these?

The output looks very similar for all these objects, so I'm wondering if we should extract the method to a helper class. I don't feel too strongly about this, so as is is fine as well.

fahia: How's moving it to the base exporter class?

lib/shopify_transporter/exporters/magento/soap.rb
fahia: Sure!

lib/shopify_transporter/exporters/magento/customer_exporter.rb
surajreddy: I'm okay with that - would we lose functionality that way?

fahia: Yeah, it'd end up being less friendly. I went ahead and tried that out but realized it actually worsens the experience. I prefer this slightly less DRY approach that can easily be changed later /shrug.

JordanVerasamy: agree w fahia

fahia:
Rubocop fails because of the begin/rescue issue. Tests are all green

spec/shopify_transporter/exporters/magento/soap_spec.rb
```diff
@@ -186,6 +186,36 @@ def expected_batching_filter(batch_key, batch_range_string)
...
+                'customer_id',
+                '3,4,5'
+              )) do
+              raise Savon::Error, 'Soap call failed.'
+            end
+
+
+            expect { |b| soap_client.call_in_batches(method: :test_call, batch_index_column: 'customer_id').each(&b) }
+              .to yield_successive_args(['batch-0,1,2'], ['batch-6,7'])
+          end
```
surajreddy: We should add a check here for the message output to $stderr as well, similar to the other tests.

fahia: good call, forgot about this

PR #86, opened 2018-10-12
Fix rubocop errors

fahia:
### What it does and why:
- Rubocop is incorrectly trying to correct a begin/rescue block which leads to a syntax error. This PR ignores the spots it complains about in order for everything to remain 🍏 
### Checklist:

### DIFF
--- a/.rubocop-https---shopify-github-io-ruby-style-guide-rubocop-yml
+++ b/.rubocop-https---shopify-github-io-ruby-style-guide-rubocop-yml
@@ -829,7 +829,7 @@ Style/Proc:
   Enabled: true
 
 Style/RedundantBegin:
-  Enabled: true
+  Enabled: false
 
 Style/RedundantException:
   Enabled: true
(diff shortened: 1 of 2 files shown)


lib/shopify_transporter/exporters/magento/customer_exporter.rb
```diff
@@ -15,6 +15,7 @@ def key
 
         def export
           base_customers.each do |customer|
+            # rubocop:disable RedundantBegin
```
surajreddy: I think we need to add an enable after the block/lines of code we care about as well, otherwise cop will skip the rest of the file.

How do you feel about just removing the `RedundantBegin` check from our rules instead? I prefer this approach.

fahia: I went ahead and did the fix. On second thought, I don't think it's terribly concerning to not have rubocop yell at us if we have redundant begin block.

fahia:
I'm gonna go ahead and merge this because it's failing on other branches and has no impact on usability..

surajreddy:
A bit late to the convo, but I'm happy with how this turned out!

Issue #87, opened 2018-10-12
Add more supports for line_item fields


zliu1020:
Fields that map to Magento but haven't been implemented:
- [x] compare_at_price
- [x] total_discount
- [x] requires_shipping
- [x] taxable 
- [x] fulfillment_status

zliu1020:
Update:
- `total_discount` is available for POS only and will be deprecated soon. 
- `fulfillable_quantity` is calculated and set by core API itself. 
We will not support those two fields. There is no way to set discount on a line item level. To set discount on an order level, use `discount_codes` instead.

PR #89, opened 2018-10-15
Add support for converting order line item attributes

zliu1020:
### What it does and why:
see https://github.com/Shopify/shopify_transporter/issues/87

### Checklist:

### demo:
screen shot 2018-10-18 at 10 25 34 am

---

### Related issues: https://github.com/Shopify/shopify_transporter/issues/87
Closes:


lib/shopify_transporter/pipeline/magento/order/line_items.rb
```diff
@@ -52,6 +57,42 @@ def tax_lines(item)
...
+          def fulfillable_quantity(item)
+            qty_ordered = qty_by_status(item, 'ordered')
+            qty_shipped = qty_by_status(item, 'shipped')
+            qty_refunded = qty_by_status(item, 'refunded')
+
+            qty_ordered - [qty_shipped, qty_refunded].max
+          end
+
+          def requires_shipping?(item)
+            item['is_virtual'].to_i == 0 ? true : false
```
JordanVerasamy: In general, any predicate

```
X ? true : false
```

will always evaluate to the same value as simply 

```
X
```

In this case, this means you can change this method body to simply be:

```
item['is_virtual'].to_i == 0
```

spec/shopify_transporter/pipeline/magento/order/line_items_spec.rb
```diff
@@ -26,19 +26,22 @@ module ShopifyTransporter::Pipeline::Magento::Order
...
-            qty_ordered: "qty ordered #{i}",
+            qty_ordered: "5",
+            qty_shipped: "5",
             sku: "sku #{i}",
             name: "name #{i}",
             price: "price #{i}",
             tax_amount: "10",
             tax_percent: "12",
+            discount_amount: "20",
+            is_virtual: "0"
```
JordanVerasamy: nit: i believe you haven't run rubocop on this, since it would swap it to single-quotes? unsure

lib/shopify_transporter/pipeline/magento/order/line_items.rb
```diff
@@ -52,6 +57,42 @@ def tax_lines(item)
...
+              'partial'
+            end
+          end
+
+          def fulfillable_quantity(item)
+            qty_ordered = qty_by_status(item, 'ordered')
+            qty_shipped = qty_by_status(item, 'shipped')
+            qty_refunded = qty_by_status(item, 'refunded')
+
+            qty_ordered - [qty_shipped, qty_refunded].max
```
JordanVerasamy: what's the logic for this line? it doesnt seem obviously correct to me

lib/shopify_transporter/pipeline/magento/order/line_items.rb
```diff
@@ -52,6 +57,42 @@ def tax_lines(item)
             ]
           end
 
+          def total_discount(item)
+            item['discount_amount'] if item['discount_amount'].present? && item['discount_amount'].to_f != 0
```
JordanVerasamy: Does this actually ever work? (followup to our IRL discussion yesterday afternoon)

We shouldn't merge this if it never does anything. We should experiment with the behaviour of this column/spike into it.

lib/shopify_transporter/pipeline/magento/order/line_items.rb
zliu1020: Good point! I'll update the code.

lib/shopify_transporter/pipeline/magento/order/line_items.rb
zliu1020: All the discount-related fields are not working as of now. I've implemented the code on the app side to prepare and send discount data, but the Shopify API is setting ``total_discounts``(top level), ``total_discount``(line item level) and ``discount_amount`` (top level) all to 0 for some reason. 

I had a conversion with @surajreddy offline. He stated that he was aware of such behaviour of the Shopify API, the code related to ``total_discounts`` in the app existed but then was removed for this reason. I'd like to get more context from Suraj after he is free from the delete incident. For now I will hold off on any implementation related to discount stuff.

lib/shopify_transporter/pipeline/magento/order/line_items.rb
JordanVerasamy: interesting nice

zliu1020:
Update:
- We will not support those two fields:
  - `total_discount` is available for POS only and will be deprecated soon. 
  - `fulfillable_quantity` is calculated and set by core API itself. 

- There is no way to set discount on a line item level. To set discount on an order level, use `discount_codes` instead.

spec/factories/shopify/order.rb
```diff
@@ -98,12 +98,12 @@
...
     sequence(:quantity) { |n| "quantity-#{n}" }
     sequence(:price) { |n| "price-#{n}" }
-    sequence(:compare_at_price) { |n| "compare-at-price-#{n}" }
     sequence(:taxable) { |n| "taxable-#{n}" }
     sequence(:requires_shipping) { |n| "requires-shipping#{n}" }
     sequence(:sku) { |n| "sku-#{n}" }
     sequence(:fulfillment_status) { |n| "fulfillment-status#{n}" }
-    sequence(:discount) { |n| "discount-#{n}" }
+    sequence(:fulfillable_quantity) { |n| "fulfillable_quantity-#{n}" }
+    sequence(:total_discount) { |n| "total_discount-#{n}" }
```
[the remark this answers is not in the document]
zliu1020: You're right! I'll remove them.

spec/files/transporter_csv_templates/orders.csv
```diff
@@ -1,4 +1,4 @@
...
+Name,Email,Financial Status,Fulfillment Status,Currency,Buyer Accepts Marketing,Cancel Reason,Cancelled At,Closed At,Tags,Note,Phone,Referring Site,Processed At,Source name,Total discounts,Total weight,Total Tax,Shipping Company,Shipping Name,Shipping Phone,Shipping First Name,Shipping Last Name,Shipping Address1,Shipping Address2,Shipping City,Shipping Province,Shipping Province Code,Shipping Zip,Shipping Country,Shipping Country Code,Billing Company,Billing Name,Billing Phone,Billing First Name,Billing Last Name,Billing Address1,Billing Address2,Billing City,Billing Province,Billing Province Code,Billing Zip,Billing Country,Billing Country Code,Lineitem name,Lineitem quantity,Lineitem price,Lineitem sku,Lineitem requires shipping,Lineitem taxable,Lineitem fulfillment status,Tax 1 Title,Tax 1 Price,Tax 1 Rate,Tax 2 Title,Tax 2 Price,Tax 2 Rate,Tax 3 Title,Tax 3 Price,Tax 3 Rate,Transaction amount,Transaction kind,Transaction status,Metafield Namespace,Metafield Key,Metafield Value,Metafield Value Type
```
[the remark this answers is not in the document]
zliu1020: Good catch. Will do.

spec/shopify_transporter/pipeline/magento/order/line_items_spec.rb
```diff
@@ -48,36 +50,42 @@ module ShopifyTransporter::Pipeline::Magento::Order
...
             price: "price",
-            tax_lines: nil
+            tax_lines: nil,
+            requires_shipping: true,
+            fulfillment_status: 'partial',
+            taxable: false
           }.deep_stringify_keys
         ]
-        expect(shopify_order['line_items']).to match_array(expected_shopify_order_line_item)
+        expect(shopify_order['line_items']).to eq(expected_shopify_order_line_item)
```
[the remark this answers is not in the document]
zliu1020: The test case for fulfillment_status is nil is in.

Issue #91, opened 2018-10-16
If the order contains a simple product, its parent product will occupy a row in the conversion CSV too


zliu1020:
Simple product in order will occupy two lines as their parent product shows up too. Shopify will treat them as two products instead of one. The solution is to remove the parent product from CSV.

In Magento:
![screen shot 2018-10-16 at 2 24 18 pm](https://user-images.githubusercontent.com/21313470/47038512-abd0a780-d14f-11e8-9e8e-32ecf951e840.png)
In CSV:
<img width="759" alt="screen shot 2018-10-16 at 2 24 34 pm" src="https://user-images.githubusercontent.com/21313470/47038497-a3786c80-d14f-11e8-927d-e6cee4505a85.png">

JordanVerasamy:
Update: We decided to handle this in the transformation step instead of the extraction step. Closed the previous PRs and implementing some similar logic in the order attributes accumulator instead.

PR #92, opened 2018-10-17
Fix a bug where child product in order CSV will occupy two lines as their parent product shows up too

zliu1020:
### What it does and why:
see https://github.com/Shopify/shopify_transporter/issues/91

### Checklist:

#### Before:
screen shot 2018-10-17 at 9 44 59 am

#### After:

screen shot 2018-10-17 at 9 47 17 am

---

### Related issues: https://github.com/Shopify/shopify_transporter/issues/91
Closes:

### DIFF
--- a/lib/shopify_transporter/exporters/magento/order_exporter.rb
+++ b/lib/shopify_transporter/exporters/magento/order_exporter.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
-
+require 'pry'
 module ShopifyTransporter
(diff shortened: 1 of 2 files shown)


JordanVerasamy:
Here's a proposal for how to do this without mutation: 

https://github.com/Shopify/shopify_transporter/pull/93

cc @zliu1020

JordanVerasamy:
Looks like we can do this in the convert step.

PR #93, opened 2018-10-17
refactor #92 to work without mutation

JordanVerasamy:
### What it does and why:

this is a possible alternate way to approach https://github.com/Shopify/shopify_transporter/pull/92 that (IMO) is more ruby-ic

### Demo

Before:
screen shot 2018-10-17 at 9 40 15 am
After:
screen shot 2018-10-17 at 9 39 51 am

^ demo images courtesy of Ziyan

### Checklist:

---

### Related issues:
Closes:
https://github.com/Shopify/shopify_transporter/issues/91

### DIFF
--- a/lib/shopify_transporter/exporters/magento/order_exporter.rb
+++ b/lib/shopify_transporter/exporters/magento/order_exporter.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
-
+require 'pry'
 module ShopifyTransporter
   module Exporters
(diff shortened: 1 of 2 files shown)


JordanVerasamy:
looks like it can be done in the convert step. i think both #92 and #93 should be closed.

cc @zliu1020

JordanVerasamy:
Looks like we can do this in the convert step.

PR #94, opened 2018-10-17
Fix default address bug

zliu1020:
### What it does and why:
See https://github.com/Shopify/transporter_app/issues/1346

### Checklist:

##### On master:
screen shot 2018-10-26 at 11 57 28 am

#### On this branch:
screen shot 2018-10-26 at 11 57 38 am

##### On master && On this branch:
screen shot 2018-10-26 at 1 00 27 pm

---

### Related issues: https://github.com/Shopify/transporter_app/issues/1346
Closes:

### DIFF
--- a/lib/shopify_transporter/shopify/customer.rb
+++ b/lib/shopify_transporter/shopify/customer.rb
@@ -2,7 +2,7 @@
 require_relative 'record'
 require 'active_support/inflector'
 require 'csv'
-
+require 'pry'
 module ShopifyTransporter
   module Shopify
     class Customer < Record
@@ -57,12 +57,21 @@ module ShopifyTransporter
       ).freeze
 
       def top_level_row_values
-        base_hash.merge(record_hash.slice(*TOP_LEVEL_ATTRIBUTES)).values
+        base_hash.merge(record_hash.slice(*TOP_LEVEL_ATTRIBUTES))
+          .merge(default_address_without_name_and_phone)
+          .values
+      end
+
+      def default_address_without_name_and_phone
+        return {} unless record_hash.dig('addresses').present?
+        default_address = record_hash['addresses'][0]
+        %w(first_name last_name phone).each { |key| default_address.delete(key) }
+        default_address
       end
 
       def address_row_values
         return [] unless record_hash['addresses']
-        record_hash['addresses'].map do |address_hash|
+        record_hash['addresses'].drop(1).map do |address_hash|
           address = address_hash.slice(*ADDRESS_ATTRIBUTES)
           populate_missing_address_attributes!(address)
           row_values_from(address) if self.class.has_values?(address)
--- a/lib/shopify_transporter/shopify/customer.rb
+++ b/lib/shopify_transporter/shopify/customer.rb
@@ -2,7 +2,7 @@
 require_relative 'record'
 require 'active_support/inflector'
 require 'csv'
-require 'pry'
+
 module ShopifyTransporter
   module Shopify
     class Customer < Record
--- a/lib/shopify_transporter/shopify/customer.rb
+++ b/lib/shopify_transporter/shopify/customer.rb
@@ -67,19 +67,17 @@ module ShopifyTransporter
       end
 
       def default_address
-        if at_least_one_address?
-          record_hash['addresses'][0].slice(*ADDRESS_ATTRIBUTES)
-        else
-          {}
-        end
+        return {} unless at_least_one_address?
+
+        record_hash['addresses'][0].slice(*ADDRESS_ATTRIBUTES)
       end
 
       def at_least_one_address?
-        record_hash.dig('addresses').present? && record_hash.dig('addresses')[0].present?
+        record_hash['addresses'].present? && record_hash['addresses'][0].present?
       end
 
       def address_row_values
-        return [] unless record_hash['addresses'].present?
+        return [] unless record_hash['addresses'].present? && record_hash['addresses'].length > 1
         record_hash['addresses'].drop(1).map do |address_hash|
           address = address_hash.slice(*ADDRESS_ATTRIBUTES)
           populate_missing_address_attributes!(address)
--- a/lib/shopify_transporter/shopify/customer.rb
+++ b/lib/shopify_transporter/shopify/customer.rb
@@ -57,20 +57,29 @@ module ShopifyTransporter
       ).freeze
 
       def top_level_row_values
-        base_hash.merge(record_hash.slice(*TOP_LEVEL_ATTRIBUTES))
-          .merge(default_address_without_name_and_phone)
-          .values
+        default_address.merge(top_level_attributes).values_at(*Customer.columns)
       end
 
-      def default_address_without_name_and_phone
-        return {} unless record_hash.dig('addresses').present?
-        default_address = record_hash['addresses'][0]
-        %w(first_name last_name phone).each { |key| default_address.delete(key) }
-        default_address
+      def top_level_attributes
+        base_hash
+          .merge(record_hash.slice(*TOP_LEVEL_ATTRIBUTES))
+          .delete_if { |_, v| v.nil? }
+      end
+
+      def default_address
+        if at_least_one_address?
+          record_hash['addresses'][0].slice(*ADDRESS_ATTRIBUTES)
+        else
+          {}
+        end
+      end
+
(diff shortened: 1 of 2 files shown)


lib/shopify_transporter/shopify/customer.rb
```diff
@@ -57,12 +57,30 @@ def to_csv
...
+      end
+
+      def top_level_attributes
+        base_hash
+          .merge(record_hash.slice(*TOP_LEVEL_ATTRIBUTES))
+          .delete_if { |_, v| v.nil? }
+      end
+
+      def default_address
+        if at_least_one_address?
```
fahia: I'd write this slightly differently, something like:

```ruby
def default_address
  return {} unless at_lease_one_address?

  record_hash['addresses'][0].slice(*ADDRESS_ATTRIBUTES)
end

File: lib/shopify_transporter/shopify/customer.rb [hunk 2]
```diff
@@ -57,12 +57,30 @@ def to_csv
...
+          record_hash['addresses'][0].slice(*ADDRESS_ATTRIBUTES)
+        else
+          {}
+        end
+      end
+
+      def at_least_one_address?
+        record_hash.dig('addresses').present? && record_hash.dig('addresses')[0].present?
```
fahia: No need for the dig since it's one level deep access. If you were doing something like `record_hash.dig('addresses', 'something_else'), then dig is necessary.

```ruby 
record_hash['addresses'].present? && record_hash['addresses'].first.present?

File: lib/shopify_transporter/shopify/customer.rb [hunk 3]
```diff
@@ -57,12 +57,30 @@ def to_csv
...
       end
 
       def address_row_values
-        return [] unless record_hash['addresses']
-        record_hash['addresses'].map do |address_hash|
+        return [] unless record_hash['addresses'].present?
+        record_hash['addresses'].drop(1).map do |address_hash|
```
fahia: The `drop(1)` feels like a hack. We should probably make sure it's the right amount of addresses before we map over it.

lib/shopify_transporter/shopify/customer.rb
surajreddy: 👍 to this. We use this pattern a lot as a guard clause.

lib/shopify_transporter/shopify/customer.rb
zliu1020: I've added check in the beginning to make sure the array is of size 2 at least. Does this make `drop(1)`seem less like a hack?

zliu1020:
@fahia The reason that the error doesn't happen on master is because `phone` in top level attributes has never been set by the code. The customers are imported with `phone` being `nil` and phone number validation has never been triggered. I've tested some rejected phone number using `Shopify-cli` and those numbers remain to be rejected. However, those customers can be saved successfully if I change the number to `nil` again. I believe the code on this branch is behaving correctly as the rejected phone numbers are indeed invalid ones.

fahia:
@zliu1020 I believe that anything in master should just work at any given time when using the transporter app. This project should encompass the fundamental requirements to get certain object types into Shopify and should _never_ fail. Anything that requires customization or further pipeline stages should be done out of the default pipeline stages. Default things should by definition work all the time.

If we were to merge this, no one could import their customers because they now have to figure out why things are failing and work backward. I propose we do one of two things:

1) Maintain behavior in master and not include phone numbers
2) Include phone numbers and ensure they match the Shopify API format by adding the functionality to do so.

lib/shopify_transporter/shopify/customer.rb
surajreddy: Yes, I'm more okay with it now. 👍

lib/shopify_transporter/shopify/customer.rb
```diff
@@ -57,12 +57,28 @@ def to_csv
...
 
       def top_level_row_values
-        base_hash.merge(record_hash.slice(*TOP_LEVEL_ATTRIBUTES)).values
+        default_address.merge(top_level_attributes).values_at(*Customer.columns)
+      end
+
+      def top_level_attributes
+        base_hash
+          .merge(record_hash.slice(*TOP_LEVEL_ATTRIBUTES))
+          .delete_if { |_, v| v.nil? }
```
surajreddy: Is there a specific reason to add `.delete_if { |_, v| v.nil? }`? If it doesn't add an obvious benefit to the csv output, let's remove it.

Otherwise, I suggest adding a test case for this.

lib/shopify_transporter/shopify/customer.rb
```diff
@@ -57,12 +57,28 @@ def to_csv
...
+      end
+
+      def default_address
+        return {} unless at_least_one_address?
+
+        record_hash['addresses'][0].slice(*ADDRESS_ATTRIBUTES)
+      end
+
+      def at_least_one_address?
```
surajreddy: 👍

lib/shopify_transporter/shopify/customer.rb
```diff
@@ -57,12 +57,28 @@ def to_csv
...
+        record_hash['addresses'].present? && record_hash['addresses'][0].present?
       end
 
       def address_row_values
-        return [] unless record_hash['addresses']
-        record_hash['addresses'].map do |address_hash|
+        return [] unless record_hash['addresses'].present? && record_hash['addresses'].length > 1
```
surajreddy: I really like `at_least_one_address?` above. How do you feel about moving this to `more_than_one_address?`?

lib/shopify_transporter/shopify/customer.rb
zliu1020: Yep there's a reason for that. 

In function `top_level_row_values`, I'm merging `top_level_attributes` into `default_address`. Before the merge happens, this `delete_if` makes sure that all `nil` values of `top_level_attributes` are removed. Otherwise, the `nil` address fields from `top_level_attributes` will overwrite existing values in `default_address`. 

The reason why I'm merging `top_level_attributes` into `default_address`, instead of the other direction around, is that I want to make sure fields like `name` and `email`in `top_level_attributes` take precedence over fields in `default_address`.

I believe `delete_if` is already covered by existing test cases as suggested by coverage report.

lib/shopify_transporter/shopify/customer.rb
zliu1020: Will do!

zliu1020:
I've noticed another issue related to default addresses in conversion step, here's a separate issue for [it ](https://github.com/Shopify/shopify_transporter/issues/106)

lib/shopify_transporter/shopify/customer.rb
```diff
@@ -57,12 +57,36 @@ def to_csv
...
+        return {} unless at_least_one_address?
+
+        record_hash['addresses'][0].slice(*ADDRESS_ATTRIBUTES)
+      end
+
+      def at_least_one_address?
+        record_hash['addresses'].present? && record_hash['addresses'][0].present?
+      end
+
+      def not_able_to_merge_address?
```
JordanVerasamy: functions like this should be written in the positive case, like `able_to_merge_address`, whenever possible. `not_` is a smell that you should be using logic operators in the code, like `!`, instead.

zliu1020: Fixed in the latest commit

lib/shopify_transporter/shopify/customer.rb
```diff
@@ -57,12 +57,36 @@ def to_csv
...
+        return {} unless at_least_one_address?
+
+        record_hash['addresses'][0].slice(*ADDRESS_ATTRIBUTES)
+      end
+
+      def at_least_one_address?
+        record_hash['addresses'].present? && record_hash['addresses'][0].present?
+      end
+
+      def able_to_merge_address?
```
surajreddy: I suggest being more specific with the method name: `able_to_merge_default_address?`

spec/shopify_transporter/shopify/customer_spec.rb
```diff
@@ -69,7 +73,21 @@ module Shopify
           )
         end
 
-        it 'produces rows with overridden last_name/first_name/phone those values exist in the address' do
+        it 'should merge the default address into the top level row when there is no conflict in doing so' do
```
surajreddy: 👍

Issue #82, opened 2018-10-10
Discounts should be converted to discount_code, amount and type instead of being output as total_discounts


surajreddy:
Discounts from Magento should be extracted to the discount fields supported by the Transporter App - https://help.shopify.com/en/manual/migrating-to-shopify/transporter-app/csv-orders :

- discount_code
- discount_amount
- discount_type

Currently they are not being exported.

PR #96, opened 2018-10-18
Handle order discount codes

zliu1020:
### What it does and why:
See https://github.com/Shopify/shopify_transporter/issues/82
### Checklist:

#### Magento:
screen shot 2018-10-19 at 3 03 23 pm
#### Shopify:
screen shot 2018-10-19 at 3 03 17 pm

---

### Related issues: https://github.com/Shopify/shopify_transporter/issues/82
Closes:


lib/shopify_transporter/pipeline/magento/order/discounts.rb
```diff
@@ -0,0 +1,103 @@
...
+
+          def discount_code(hash)
+            hash['discount_description'].present? ? hash['discount_description'] : 'Magento'
+          end
+
+          def discount_amount(hash)
+            hash['discount_amount'].present? ? hash['discount_amount'][1..-1].to_f : 0
+          end
+
+          def fetch_value(hash, key)
```
JordanVerasamy: potentially rename this to `value_as_float`? i dont like the fact that we're converting the value we read without having that be evident in the name

lib/shopify_transporter/pipeline/magento/order/discounts.rb
```diff
@@ -0,0 +1,103 @@
...
+              {
+                code: discount_code(hash),
+                amount: discount_percentage,
+                type: 'percentage',
+              }.stringify_keys
+            end
+          end
+
+          def discount_percentage(hash)
+            percentage = 0
```
JordanVerasamy: IMO, any time you write a function that looks like this:

```
def foo(a, b, c, ...)
  percentage = 0
  ...
  percentage
end
```

that's a smell, and it should signal to you that maybe there's a better way to do the same thing.

In this case, I'm not sure what the best way forward is, but here's my first gut reaction on how to write this function:

```
def discount_percentage(hash)
  return 0 unless hash.dig('items', 'result', 'items', 'item').present?

  if line_items(hash).is_a?(Hash)
    fetch_value(line_items(hash), 'discount_percent')
  else
    discounts = line_items(hash).map do |line_item|
      fetch_value(line_item, 'discount_percent')
    end.uniq
    discount_applied_on_all_line_items?(discounts) ? discounts.first : 0
  end
end
```

Any reason this wouldn't work?

JordanVerasamy: Also, maybe people actually find this less clear -- I personally prefer it, but interested in other people's opinions as well.

JordanVerasamy: Also honestly I would prefer just checking `discounts.length == 1` rather than having `discount_applied_on_all_line_items?` which is really just an unnecessary layer of indirection. Again, not sure about how other people feel about this.

lib/shopify_transporter/pipeline/magento/order/discounts.rb
```diff
@@ -0,0 +1,103 @@
...
+            end
+            percentage
+          end
+
+          def discount_code(hash)
+            hash['discount_description'].present? ? hash['discount_description'] : 'Magento'
+          end
+
+          def discount_amount(hash)
+            hash['discount_amount'].present? ? hash['discount_amount'][1..-1].to_f : 0
```
JordanVerasamy: As discussed IRL, the purpose of `[1..-1]` isn't clear -- apparently it's being used to remove the leading `-` from Magento discounts, since Magento uses negative numbers for discounts but Shopify expects positive numbers.

I'd prefer something more explicit about its purpose, such as taking the absolute value of the converted float instead of manipulating the string itself.

lib/shopify_transporter/pipeline/magento/order/discounts.rb
zliu1020: Good point!

lib/shopify_transporter/pipeline/magento/order/discounts.rb
zliu1020: @JordanVerasamy Could u elaborate why this code style is considered as smelly code? I don't think I completely follow.

zliu1020: I extracted the check of `discounts.length == 1` into a separate method because the name of the method explains the reason why this check is necessary. I'm open to others' opinions on this matter.

lib/shopify_transporter/pipeline/magento/order/discounts.rb
zliu1020: Will do! Absolute values sounds reasonable to me.

lib/shopify_transporter/pipeline/magento/order/discounts.rb
```diff
@@ -0,0 +1,103 @@
...
+              percentage_discount(hash),
+            ].compact
+          end
+
+          def shipping_discount(hash)
+            shipping_discount_amount = fetch_value(hash, 'shipping_discount_amount')
+            shipping_amount = fetch_value(hash, 'shipping_amount')
+            if shipping_discount_amount > shipping_amount
+              {
+                code: discount_code(hash),
```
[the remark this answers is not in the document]
zliu1020: `hash['discount_description']` is not guaranteed to be present but the `discount_code` is a required field for Shopify API. That's why we need a default value for it.

Personally I find just `Magento` is more concise than `Magento blah discount`. The `discount_type` field is part of the `discount` object anyways. I'm open to a second opinion if someone else wants to share their thoughts on this matter.

lib/shopify_transporter/pipeline/magento/order/discounts.rb
```diff
@@ -0,0 +1,103 @@
...
+                amount: shipping_discount_amount,
+                type: 'shipping',
+              }.stringify_keys
+            end
+          end
+
+          def fixed_amount_discount(hash)
+            discount_amount = discount_amount(hash)
+            discount_percentage = discount_percentage(hash)
+            if discount_amount > 0 && discount_percentage == 0
```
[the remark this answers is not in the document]
zliu1020: As far as I can tell from our Magento sample data, they're mutually exclusive.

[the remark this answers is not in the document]
zliu1020: I believe it's possible for line items to have different `fixed`/`percentage` discounts, however, we don't support setting discounts on a line item level using Shopify API. This situation is recognized and handled as a `fixed_amount` discount applied on an order level.

zliu1020:
In Shopify the shipping discounts work in the way that users either receive free shipping or have to pay all the costs. 
According to [this link ](https://magento.stackexchange.com/questions/52886/how-can-i-give-discount-on-shipping-costs), Magento does not support partial shipping discounts either. There is hack to achieve the same results but it is not supported by default. 
Therefore, the conversion code won't bother implementing logic for partial shipping discounts.

spec/shopify_transporter/pipeline/magento/order/discounts_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+module ShopifyTransporter::Pipeline::Magento::Order
+  RSpec.describe Discounts, type: :helper do
+    context '#convert' do
+      it "extract fixed amount discount into correct discount codes in Shopify" do
+        magento_order = FactoryBot.build(:magento_order, :with_fixed_amount_discount)
+        shopify_order = described_class.new.convert(magento_order, {})
+        expected_discount_code = [
+          {
+            amount: 15,
+            code: 'TEST_DISCOUNT_CODE',
```
JordanVerasamy: Personally I'd replace `'TEST_DISCOUNT_CODE'` with `magento_order['discount_description']` or whatever, because that way if you change the factory you dont need to also change the tests to match.

There's definitely an argument for keeping simpler tests though, and this is the nittiest of picks. Do whatever you think is better, this is non-blocking.

lib/shopify_transporter/pipeline/magento/order/discounts.rb
```diff
@@ -0,0 +1,101 @@
...
+            return 0 unless hash.dig('items', 'result', 'items', 'item').present?
+
+            if line_items(hash).is_a?(Hash)
+              value_as_float(line_items(hash), 'discount_percent')
+            else
+              discounts = line_items(hash).map do |line_item|
+                value_as_float(line_item, 'discount_percent')
+              end.uniq
+              discount_applied_on_all_line_items?(discounts) ? discounts.first : 0
+            end
```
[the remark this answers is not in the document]
JordanVerasamy: this would be fine, i think the current implementation is fine too though -- especially if the if/elsif construct was swapped for a case like so:

```
case line_items(hash)
when Hash
...
when Array
...
end
```

this works without the need for any `is_a?` calls, because case/when constructs automatically match using `===`, and `Hash === {}`.

(unintuitively, though, `{} === Hash` is not true. the `===` operator is not symmetric for some reason.

```
irb(main):003:0> Hash === {}
=> true
irb(main):004:0> Hash == {}
=> false
```

ask me about this if you're curious, it's one of the things that annoys me about Ruby

JordanVerasamy:
🎩

lib/shopify_transporter/pipeline/magento/order/discounts.rb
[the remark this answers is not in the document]
surajreddy: We chatted about this offline - I missed that these also had a `type` key indicating the type.

lib/shopify_transporter/pipeline/magento/order/discounts.rb
```diff
@@ -0,0 +1,115 @@
...
+
+          def line_items?(hash)
+            hash.dig('items', 'result', 'items', 'item').present?
+          end
+
+          def line_items(hash)
+            hash['items']['result']['items']['item']
+          end
+
+          def all_discount_percentages(hash)
```
surajreddy: If you're worried about performance, one optimization you can do here is:
```
def all_discount_percentages(hash)
  @all_discount_percentages ||= ...
end
```

This will store the value in the instance variable all_discount_percentages the first time it's computed, and reuse it the second time.

I don't think it's needed here.

spec/factories/shopify/order.rb
```diff
@@ -77,6 +77,23 @@
...
+      discounts do
+        [
+          {
+            'amount': 25,
+            'code': '25OFF',
+            'type': 'fixed_amount'
+          }.stringify_keys,
+          {
+            'amount': 20,
+            'kind': 'FREESHIPPINGUNDER20',
```
surajreddy: `code`?

lib/shopify_transporter/shopify/order.rb
```diff
@@ -65,7 +67,7 @@ def to_csv
       LINE_ITEM_PREFIX = 'lineitem_'
 
       LINE_ITEM_ATTRIBUTES = %w(
-        name quantity price sku requires_shipping taxable fulfillment_status
+        name quantity price discount compare_at_price sku requires_shipping taxable fulfillment_status
```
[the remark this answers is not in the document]
surajreddy: The transporter sample csv has compare_at_price, so I think this is fine.

If we decide to leave it, we should update the spec to have this column as well.

Can you double check that the tests are correct for this? The headers are different from the test and they're still passing - I think it's because they have the same number of columns.

Issue #88, opened 2018-10-15
Unpaid Magento orders show up as being paid in Shopify


zliu1020:
To reproduce, use Order # 100000081 in Magento. Extract, convert and load this order into Shopify.

In Magento, the status is unpaid
![screen shot 2018-10-15 at 2 54 13 pm](https://user-images.githubusercontent.com/21313470/46971585-639b8180-d08a-11e8-90eb-c1e4d189c9a6.png)

In Shopify, the order is fully paid
![screen shot 2018-10-15 at 2 55 20 pm](https://user-images.githubusercontent.com/21313470/46971595-6c8c5300-d08a-11e8-9220-30fa05759b65.png)

zliu1020:
Update: We've got in touch with core team and decided on setting unpaid order to `pending`.

PR #98, opened 2018-10-19
Fix unpaid order status bug

zliu1020:
### What it does and why:
see https://github.com/Shopify/shopify_transporter/issues/88

### Checklist:

---

### Related issues: https://github.com/Shopify/shopify_transporter/issues/88
Closes:

### DIFF
--- a/lib/shopify_transporter/pipeline/magento/order/top_level_attributes.rb
+++ b/lib/shopify_transporter/pipeline/magento/order/top_level_attributes.rb
@@ -46,7 +46,7 @@ module ShopifyTransporter
             order_state = hash['state']
             status = nil
 
-            if order_state == 'Pending Payment'
+            if order_state == 'Pending Payment' || unpaid?(hash)
               status = 'pending'
             elsif paid?(hash)
               status = 'paid'
@@ -105,6 +105,10 @@ module ShopifyTransporter
 
           def paid?(hash)
             total_price(hash) == total_paid(hash) && total_refunded(hash) == 0
+          end
+
+          def unpaid?(hash)
+            total_paid(hash) == 0
           end
 
           def partially_paid?(hash)
--- a/lib/shopify_transporter/pipeline/magento/order/top_level_attributes.rb
+++ b/lib/shopify_transporter/pipeline/magento/order/top_level_attributes.rb
@@ -43,21 +43,17 @@ module ShopifyTransporter
           end
 
           def financial_status(hash)
-            order_state = hash['state']
-            status = nil
-
-            if order_state == 'Pending Payment' || unpaid?(hash)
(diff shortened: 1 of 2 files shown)


lib/shopify_transporter/pipeline/magento/order/top_level_attributes.rb
```diff
@@ -46,7 +46,7 @@ def financial_status(hash)
             order_state = hash['state']
             status = nil
 
-            if order_state == 'Pending Payment'
+            if order_state == 'Pending Payment' || unpaid?(hash)
```
surajreddy: One side note: is it possible for an order to have total_paid = 0 and have refunds?

surajreddy: Since 'pending' is the default state for orders in Shopify, should we make this the default on like 47?

Instead of `status = nil`, we can set `status = 'pending'`, and remove this if statement.

How do you feel about this?

zliu1020: hmm I doubt that's possible. There's nothing to refund for if the user has paid nothing in the first place.

zliu1020: Good call! I'll update the code.

lib/shopify_transporter/pipeline/magento/order/top_level_attributes.rb
```diff
@@ -43,21 +43,17 @@ def build_customer(hash)
...
 
           def financial_status(hash)
-            order_state = hash['state']
-            status = nil
-
-            if order_state == 'Pending Payment'
-              status = 'pending'
-            elsif paid?(hash)
-              status = 'paid'
+            if paid?(hash)
```
surajreddy: 👏

lib/shopify_transporter/pipeline/magento/order/top_level_attributes.rb
```diff
@@ -104,7 +100,11 @@ def total_refunded(hash)
           end
 
           def paid?(hash)
-            total_price(hash) == total_paid(hash) && total_refunded(hash) == 0
+            total_price(hash) == total_paid(hash) && total_paid(hash) > 0 && total_refunded(hash) == 0
+          end
+
+          def unpaid?(hash)
```
surajreddy: This is no longer used and we can remove it.

JordanVerasamy: ^ 👍

zliu1020: Will do

PR #101, opened 2018-10-23
[WIP] fix behaviour when simple product has no parent id

JordanVerasamy:
### Checklist:

---

### Related issues:
Closes:

### DIFF
--- a/lib/shopify_transporter/pipeline/magento/product/top_level_variant_attributes.rb
+++ b/lib/shopify_transporter/pipeline/magento/product/top_level_variant_attributes.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 require 'shopify_transporter/pipeline/stage'
 require 'shopify_transporter/shopify'
+require 'pry'
 
 module ShopifyTransporter
   module Pipeline
@@ -9,9 +10,9 @@ module ShopifyTransporter
         class TopLevelVariantAttributes < Pipeline::Stage
           def convert(hash, record)
             return unless input_applies?(hash)
-            simple_product_in_magento_format = record['variants'].select do |product|
+            simple_product_in_magento_format = record['variants'].find do |product|
               product['product_id'] == hash['product_id']
-            end.first
+            end
             accumulator = TopLevelVariantAttributesAccumulator.new(simple_product_in_magento_format)
             accumulator.accumulate(hash)
           end
--- a/lib/shopify_transporter/pipeline/magento/product/top_level_variant_attributes.rb
+++ b/lib/shopify_transporter/pipeline/magento/product/top_level_variant_attributes.rb
@@ -17,7 +17,7 @@ module ShopifyTransporter
           end
 
           def input_applies?(input)
-            true unless input['parent_id'].nil?
+            input['type'] == 'simple'
           end
 
           class TopLevelVariantAttributesAccumulator < Shopify::AttributesAccumulator
--- a/spec/shopify_transporter/pipeline/magento/product/top_level_variant_attributes_spec.rb
+++ b/spec/shopify_transporter/pipeline/magento/product/top_level_variant_attributes_spec.rb
@@ -6,7 +6,7 @@ module ShopifyTransporter::Pipeline::Magento::Product
   RSpec.describe TopLevelVariantAttributes, type: :helper do
     context '#convert' do
       it 'extracts top level product variant attributes from an input hash' do
-        child_product = FactoryBot.build(:advanced_magento_simple_product)
+        child_product = FactoryBot.build(:advanced_magento_simple_product, :with_parent_id)
         parent_product = FactoryBot.build(:advanced_magento_configurable_product,
           variants: [
             {
@@ -63,8 +63,8 @@ module ShopifyTransporter::Pipeline::Magento::Product
       end
 
       it 'make sure the correct product is being converted when there exist multiple simple products as variants' do
-        child_product = FactoryBot.build(:advanced_magento_simple_product)
-        another_child_product = FactoryBot.build(:advanced_magento_simple_product)
+        child_product = FactoryBot.build(:advanced_magento_simple_product, :with_parent_id)
+        another_child_product = FactoryBot.build(:advanced_magento_simple_product, :with_parent_id)
         parent_product = FactoryBot.build(:advanced_magento_configurable_product,
           variants: [
             {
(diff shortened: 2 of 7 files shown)


JordanVerasamy:
Fixed Ziyan's complaint above. Have crosschecked behaviour with master's in a number of cases to make sure the output matches what is expected.

lib/shopify_transporter/pipeline/magento/product/top_level_variant_attributes.rb
```diff
@@ -8,18 +8,15 @@ module Magento
       module Product
         class TopLevelVariantAttributes < Pipeline::Stage
           def convert(hash, record)
-            return unless input_applies?(hash)
-            simple_product_in_magento_format = record['variants'].select do |product|
+            return {} unless record.key?('variants')
+            simple_product_in_magento_format = record['variants'].find do |product|
```
JordanVerasamy: this is the most confusing part of the PR. it turns out that you need to do this instead of just directly passing `hash` to the accumulator, because the accumulator modifies what is passed in, so you specifically need to pass in the copy of the `hash` that already exists in `record` thanks to the `VariantAttributes` pipeline stage.

JordanVerasamy: if you're confused about this, well, me too. we should really really rewrite this in a way that this doesn't happen

JordanVerasamy: This is the sort of code that reminds me how annoying mutation is. For example why doesn't this work?

```
          def convert(hash, record)
            simple_product = corresponding_variant_in_record(hash, record)
            return {} unless simple_product.present?
            accumulator = TopLevelVariantAttributesAccumulator.new(simple_product)
            accumulator.accumulate(hash)
          end

          def corresponding_variant_in_record(hash, record)
            return unless record.key?('variants')
            simple_product_in_magento_format = record['variants'].find do |product|
              product['product_id'] == hash['product_id']
            end
            return unless simple_product_in_magento_format.present?
          end
```

It seems like this should be equivalent to the code as written. However, it causes all the tests to fail.

I'm assuming this has to do with Ruby's pass-reference-by-value stuff or whatever, but I don't really know. At this point I want to just touch it as little as possible and make the fix I need to make without breaking anything.

JordanVerasamy: And the fact that I feel that way is indicative of tech debt! If your code is so confusing that you feel afraid to touch it... that's real bad.

spec/shopify_transporter/pipeline/magento/product/variant_attributes_spec.rb
```diff
@@ -1,13 +1,13 @@
 # frozen_string_literal: true
 require 'shopify_transporter/pipeline/magento/product/variant_attributes'
-
+require 'pry'
```
surajreddy: We should remove the `pry` require.

spec/shopify_transporter/pipeline/magento/product/top_level_attributes_spec.rb
```diff
@@ -5,6 +5,13 @@
 module ShopifyTransporter::Pipeline::Magento::Product
   RSpec.describe TopLevelAttributes, type: :helper do
     context '#convert' do
+      it 'does not accumulate anything if product is not a top level product' do
+        magento_product = FactoryBot.build(:magento_product, :with_parent_id)
+        shopify_product = described_class.new.convert(magento_product, {})
+
+        expect(shopify_product).to eq({})
```
surajreddy: Because TopLevelAttributes mutates the input, I suggest:
```
shopify_product = {}
described_class.new.convert(magento_product, shopify_product)
expect(shopify_product).to eq({})
```

spec/shopify_transporter/pipeline/magento/product/top_level_variant_attributes_spec.rb
```diff
@@ -5,8 +5,15 @@
 module ShopifyTransporter::Pipeline::Magento::Product
   RSpec.describe TopLevelVariantAttributes, type: :helper do
     context '#convert' do
+      it 'does not accumulate anything if product is not a simple product' do
+        magento_product = FactoryBot.build(:configurable_magento_product)
+        shopify_product = described_class.new.convert(magento_product, {})
```
surajreddy: Similar to above, I suggest:
```
shopify_product = {}
described_class.new.convert(magento_product, shopify_product)
```

spec/shopify_transporter/pipeline/magento/product/variant_attributes_spec.rb
JordanVerasamy: oof

spec/shopify_transporter/pipeline/magento/product/variant_image_spec.rb
```diff
@@ -84,15 +84,22 @@ module ShopifyTransporter::Pipeline::Magento::Product
       it 'should skip variant image conversion if the child product has no image attached' do
         simple_product_without_img = FactoryBot.build(:simple_magento_product)
         configurable_product_with_img = FactoryBot.build(:configurable_magento_product, :with_img, 'variants': [simple_product_without_img])
-        should_skip_conversion = described_class.new.convert(simple_product_without_img, configurable_product_with_img).nil?
-        expect(should_skip_conversion).to be true
+        result = described_class.new.convert(simple_product_without_img, configurable_product_with_img)
+        expect(result).to eq({})
```
surajreddy: This should check for mutation:
```
described_class.new.convert(simple_product_without_img, configurable_product_with_img)
expect(configurable_product_with_img) not to have the variant image.
```

spec/shopify_transporter/pipeline/magento/product/variant_image_spec.rb
```diff
@@ -84,15 +84,22 @@ module ShopifyTransporter::Pipeline::Magento::Product
...
       end
 
-      it 'should skip variant image conversion if the input product is not a child product' do
+      it 'should skip variant image conversion if the input product is a configurable product' do
         magento_product = FactoryBot.build(:magento_product)
         configurable_product_with_img = FactoryBot.build(:configurable_magento_product, :with_img, 'variants': [magento_product])
-        should_skip_conversion = described_class.new.convert(magento_product, configurable_product_with_img).nil?
-        expect(should_skip_conversion).to be true
+        result = described_class.new.convert(magento_product, configurable_product_with_img)
+        expect(result).to eq({})
```
surajreddy: This should check for mutation on configurable_product_with_img instead.

spec/shopify_transporter/pipeline/magento/product/variant_image_spec.rb
```diff
@@ -84,15 +84,22 @@ module ShopifyTransporter::Pipeline::Magento::Product
...
+      end
+
+      it 'should skip variant image conversion if the input product does not have a parent product' do
+        magento_product = FactoryBot.build(:magento_product)
+        configurable_product_with_img = FactoryBot.build(:simple_magento_product, :with_img, 'variants': [magento_product])
+        result = described_class.new.convert(magento_product, configurable_product_with_img)
```
surajreddy: This should check for mutation on configurable_product_with_img instead.

zliu1020:
I've 🎩  the pr for both pairs of child/parent products and simple products without parent. Works perfectly!

Issue #79, opened 2018-10-10
Extract and convert order shipping_lines


zliu1020:
Shipping_line is part of the sample CSV in Transporter app doc but not supported yet by our CLI
<img width="806" alt="screen shot 2018-10-10 at 10 40 52 am" src="https://user-images.githubusercontent.com/21313470/46744750-cc918c80-cc79-11e8-866c-96cf64086ba1.png">
<img width="997" alt="screen shot 2018-10-10 at 10 40 20 am" src="https://user-images.githubusercontent.com/21313470/46744751-cc918c80-cc79-11e8-85f5-6bbccef0b5aa.png">

zliu1020:
Update: 

We've decided not to extract any tracking information from Magento, as such info for past orders don't add any value. 

For now the code will extract `Shipping line code`, `Shipping line price`, `Shipping line title`, `Shipping line carrier identifier`, and `Shipping Tax Price` and merge them as one `shipping_line`. 

It would have been wonderful if we could implement support for retrieving multiple shipping lines if an order has that many shipments. Unluckily the limited sample data we have stopped me from further investigation. 

The current implementation will enable the basic shipping info and cost to show up in Shopify UI, so that the total charged amount can add up.  I believe this will be a improvement from the current behaviour (Amount not adding up at all cause shipping cost is missing) so it's worth implementing.

PR #107, opened 2018-10-26
Extract and convert order shipping_lines

zliu1020:
### What it does and why:
We've decided not to extract any tracking information from Magento, as such info for past orders don't add any value. 

For now the code will extract `Shipping line code`, `Shipping line price`, `Shipping line title`, `Shipping line carrier identifier`, and `Shipping Tax Price` and merge them as one `shipping_line`. 

It would have been wonderful if we could implement support for retrieving multiple shipping lines if an order has that many shipments. Unluckily the limited sample data we have stopped me from further investigation. 

The current implementation will enable the basic shipping info and cost to show up in Shopify UI, so that the total charged amount can add up.  I believe this will be a improvement from the current behaviour (Amount not adding up at all cause shipping cost is missing) so it's worth implementing. 

### Checklist:

#### With `order_id` = 98 in Magento:
screen shot 2018-10-29 at 1 47 43 pm
screen shot 2018-10-29 at 1 47 48 pm

#### After uploaded into Shopify:
screen shot 2018-10-29 at 1 47 04 pm

---

### Related issues: https://github.com/Shopify/shopify_transporter/issues/79
Closes:


lib/shopify_transporter/pipeline/magento/order/shipping_lines.rb
```diff
@@ -0,0 +1,45 @@
...
+              }.stringify_keys,
+            ]
+          end
+
+          def shipping_tax_lines(input)
+            return unless input['shipping_tax_amount'].present? && input['shipping_tax_amount'].to_f != 0
+
+            [
+              {
+                price: input['shipping_tax_amount'],
```
fahia: Why not just `'price': input['shipping_tax_amount']` instead and avoid the call to `stringify_keys`? I'd probably make this change throughout this file just because it seems a little roundabout if we know we expect string keys but use symbols and then only to convert them.

lib/shopify_transporter/pipeline/magento/order/shipping_lines.rb
```diff
@@ -0,0 +1,45 @@
...
+require 'shopify_transporter/pipeline/stage'
+require 'shopify_transporter/shopify'
+
+module ShopifyTransporter
+  module Pipeline
+    module Magento
+      module Order
+        class ShippingLines < Pipeline::Stage
+          def convert(input, record)
+            record.merge!(
```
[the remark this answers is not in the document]
zliu1020: Good point! `merge` works just find. I'll make this change.

spec/shopify_transporter/shopify/order_spec.rb
```diff
@@ -117,6 +127,16 @@ module Shopify
           ].join
           expect(actual_csv).to eq(expected_csv)
         end
+
+        it 'outpus shipping lines correctly' do
```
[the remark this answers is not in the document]
zliu1020: :Facepalm:

lib/shopify_transporter/pipeline/magento/order/shipping_lines.rb
zliu1020: Make sense! I did the call of `stringify_keys` just because most of pipeline stages are using it and I followed the same pattern. I'll update the code.

fahia: `'price': input['shipping_tax_amount']` is equivalent to `price: input['shipping_tax_amount']` as it's just syntactic sugar to be able to specify a string with a colon (but ruby still sets it to a symbol).

Forget this comment, I forgot about that ^

lib/shopify_transporter/shopify/order.rb
```diff
@@ -139,6 +148,28 @@ def discount_row_values
           row_values_from(discount) if self.class.has_values?(discount)
         end.compact
       end
+
+      def shipping_line_row_values
+        return [] unless record_hash['shipping_lines']
+
+        record_hash['shipping_lines'].map do |shipping_line_hash|
+          shipping_line = shipping_line_hash.slice(*SHIPPING_LINE_ATTRIBUTES)
+            .transform_keys! { |k| "#{SHIPPING_LINE_PREFIX}#{k}" }
```
JordanVerasamy: why `transform_keys!` instead of `transform_keys?`

Issue #111, opened 2018-11-02
Issues with Magento 1.9.3.4


michaelwolz:
I tried to export customers from our old magento store. All permissions are set for the SOAP user, but I am getting a NoMethodError all the time. Is there anything I am missing? 

![screenshot 2018-11-02 11 54 35](https://user-images.githubusercontent.com/7479806/47911796-b6e93e80-de96-11e8-99ac-15ebe1b7a219.png)

JordanVerasamy:
Looking into this. cc @michaelwolz

JordanVerasamy:
Hey @michaelwolz -- I merged https://github.com/Shopify/shopify_transporter/pull/112. You should try again. You'll probably still get an error, but at least it'll be a useful error message this time. Let me know what happens, post a screenshot, and I can start trying to debug the root cause.

michaelwolz:
Hey @JordanVerasamy, thank you :). I am getting the error "Extract error: Unable to obtain SOAP session ID from server."

![screenshot 2018-11-05 18 34 50](https://user-images.githubusercontent.com/7479806/48015682-fb403d00-e129-11e8-8b28-dab7e9dd58b0.png)

JordanVerasamy:
By the way @michaelwolz, I'm deleting the screenshot you uploaded above since it has a session ID plainly visible. I'm like 98% confident that this isn't actually a security issue in any way. Still, in the future, please don't include credentials or session identifiers in issues you post here. Thanks!

michaelwolz:
I can confirm this. Same error for me.

surajreddy:
Thank you for trying this again so quickly @rahulr-codal and @michaelwolz !

I suspect that this is happening because the SOAP calls your instances support are different from the soap calls we're expecting.

Would it be possible for you to upload the WSDL for your Magento instances? This is already publicly available and usually does not contain any sensitive information. It lists the SOAP calls/methods your Magento instance supports.

You can get to the WSDL by going to `https://yourmagentourl.com/api/v2_soap?wsdl` in your browser, and then use your browser's *Save page* functionality to save a copy.

We can use this file to see where the calls from your instance are different from ours.

michaelwolz:
Hello @surajreddy I attached the WSDL as txt file. Unfortunately GitHub is not supporting XML-Files in comments.

[v2_soap.txt](https://github.com/Shopify/shopify_transporter/files/2554563/v2_soap.txt)

michaelwolz:
[...]

PR #112, opened 2018-11-02
Add more detailed error messaging for login failures

JordanVerasamy:
### What it does and why:

We need more detailed information if we want to be able to debug why people can't log in properly. This is Step 1 towards fixing that issue

### Checklist:

---

### Related issues:
Closes:
https://github.com/Shopify/shopify_transporter/issues/111

### DIFF
--- a/lib/shopify_transporter/exporters/magento/soap.rb
+++ b/lib/shopify_transporter/exporters/magento/soap.rb
@@ -2,7 +2,6 @@
 
 require 'json'
 require 'savon'
-require 'pry'
 
 module ShopifyTransporter
   module Exporters
--- a/lib/shopify_transporter/exporters/magento/soap.rb
+++ b/lib/shopify_transporter/exporters/magento/soap.rb
@@ -90,9 +90,10 @@ module ShopifyTransporter
             }
           ).body
 
-          raise FailedLoginError, result.to_s unless result&.dig(:login_response, :login_return).present?
+          session_id = result&.dig(:login_response, :login_return)
+          raise FailedLoginError, result.to_s unless session_id.present?
 
-          @soap_session_id ||= result.dig(:login_response, :login_return)
+          @soap_session_id ||= session_id
         end
 
         def batching_filter(starting_id, ending_id, batch_index_column)
--- a/lib/shopify_transporter/exporters/magento/soap.rb
+++ b/lib/shopify_transporter/exporters/magento/soap.rb
@@ -10,7 +10,7 @@ module ShopifyTransporter
         class FailedLoginError < Exporters::ExportError
           def initialize(error_message)
             sep = $INPUT_RECORD_SEPARATOR
-            super("Unable to obtain SOAP session ID from server.#{sep + sep}Details:#{sep + sep + error_message}")
+            super("Unable to obtain SOAP session ID from server.#{sep + sep}Details:#{sep + error_message}")
           end
         end
 
--- a/lib/shopify_transporter/exporters/magento/soap.rb
+++ b/lib/shopify_transporter/exporters/magento/soap.rb
@@ -2,11 +2,18 @@
 
 require 'json'
 require 'savon'
+require 'pry'
 
 module ShopifyTransporter
   module Exporters
     module Magento
       class Soap
+        class FailedLoginError < Exporters::ExportError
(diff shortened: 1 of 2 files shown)


spec/shopify_transporter/exporters/magento/soap_spec.rb
```diff
@@ -43,14 +43,14 @@ def stub_login_call(mock_client)
         end
 
         describe '#call' do
-          it 'initializes savon with the right parameters' do
+          it 'raises FailedLoginError if login response does not contain a session id' do
```
fahia: do we no longer find value in a test that checks `things initialized properly with the right params`?

lib/shopify_transporter/exporters/magento/soap.rb
```diff
@@ -74,13 +80,19 @@ def soap_client
         end
 
         def soap_session_id
-          @soap_session_id ||= soap_client.call(
+          return @soap_session_id if @soap_session_id.present?
```
fahia: Nice 👍

lib/shopify_transporter/exporters/magento/soap.rb
```diff
@@ -74,13 +80,19 @@ def soap_client
...
+
+          result = soap_client.call(
             :login,
             message: {
               username: @username,
               api_key: @api_key,
             }
-          ).body[:login_response][:login_return]
+          ).body
```
fahia: Given we do the same digging twice, I think this would be preferable:
```ruby
     returned_response = result&.dig(:login_response, :login_return)
     raise FailedLoginError, result.to_s unless returned_response.present?
     @soap_session_id ||= returned_response

File: lib/shopify_transporter/exporters/magento/soap.rb [hunk 4]
```diff
@@ -7,6 +7,12 @@ module ShopifyTransporter
   module Exporters
     module Magento
       class Soap
+        class FailedLoginError < Exporters::ExportError
+          def initialize(error_message)
+            super("Unable to obtain SOAP session ID from server.#{$/ + $/}Details:#{$/ + $/ + error_message}")
```
fahia: Should we have a test showing the exception's response and how it's formatted (I believe, we have the same sorts of tests for other exceptions, outputs)

zliu1020: 👍 to Fahia's suggestion

spec/shopify_transporter/exporters/magento/soap_spec.rb
zliu1020: I think the test case where`things initialized properly with the right params` is included in the next case `creates a session correctly if login response contains a session id`. Correct me if I'm wrong.

JordanVerasamy: Since we now need the login response body to have a certain structure in order for things to initialize properly: if we were to have a test that tested that, it would have to stub the login response, which means it would look exactly like the test that starts on line 50.

JordanVerasamy: Yep, exactly what Ziyan said ^.

lib/shopify_transporter/exporters/magento/soap.rb
zliu1020: @JordanVerasamy This is non-blocking but would be great if u could fix it.

zliu1020: ^^ ignore this, github screwed me over ; _ ;

lib/shopify_transporter/exporters/magento/soap.rb
```diff
@@ -7,6 +7,13 @@ module ShopifyTransporter
   module Exporters
     module Magento
       class Soap
+        class FailedLoginError < Exporters::ExportError
+          def initialize(error_message)
+            sep = $INPUT_RECORD_SEPARATOR
+            super("Unable to obtain SOAP session ID from server.#{sep + sep}Details:#{sep + sep + error_message}")
```
surajreddy: Super Minor, I vote for one less `sep` after `Details:`

JordanVerasamy: Done!

Issue #113, opened 2018-11-05
Extract error: Unable to obtain SOAP session ID from server


rahulr-codal:
#### Problem:
Getting following output in .json file when doing extraction:
`[

]Extract error: Unable to obtain SOAP session ID from server.

Details:
{:login_response_param=>{:result=>"4bfd18cf7fca76537b122a84f7c2b284"}}`

**Ruby:  2.5.3**

I'm using the latest master branch.

michaelwolz:
Getting the exact same Error. Related to #111

JordanVerasamy:
Thanks to both of you. Continuing to investigate this.

JordanVerasamy:
Hmm. Looks like when we try to authenticate before starting the export, we expect a response of the form:

```
{:login_response=>{:login_return=>"SESSION_ID"}}
```

But for both of you, we're getting this response instead:

```
{:login_response_param=>{:result=>"SESSION_ID"}}`
```

This could be solved by adding a case to the code that parses the login response, to make sure we handle this correctly.

What versions of Magento are you two using? If there's a version mismatch, that could explain why the API responses are different.

If you want a quick and dirty solution, try cloning the repository yourself, changing the code in `lib/shopify_transporter/exporters/magento/soap.rb` lines 94-95 to match the response your APIs are giving you.

I can decide on a more permanent solution once I know the root cause of the API giving differently-structured responses upon authentication.

JordanVerasamy:
(Tagging @michaelwolz and @rahulr-codal)

JordanVerasamy:
By the way @rahulr-codal, I'm editing out the session ID you posted above. I'm like 98% confident that this isn't actually a security issue in any way. Still, in the future, please don't include credentials or session identifiers in issues you post here. Thanks!

michaelwolz:
We are running Magento Version 1.9.3.4

JordanVerasamy:
Man, that is really weird. We're on Magento 1.9.3.9 and the API definitely gives a different response. The Magento version changelog doesn't say anything about changing the response between those versions.

For now, I'm going to create a hotfix that allows you to log in properly. Longer-term I'd still like to figure out what's actually going on, though.

JordanVerasamy:
Released version 2.4.2 to Rubygems. Install using `gem install shopify_transporter`. I'm leaving this issue open until you try the fix and confirm it works.

rahulr-codal:
[...]

PR #117, opened 2018-11-05
Retrieve session id correctly from alternate login response from API

### LINKED ISSUE #111: Issues with Magento 1.9.3.4

JordanVerasamy:
### What it does and why:

Follow up PR to #112. Fixes #111 and #113.

Seems like the API somehow sometimes gives us the session ID in a weird format. I haven't found the root cause yet, but since it seems like this issue isn't exactly rare, we should implement a stopgap measure. So for now we will accept both formats.

I have no way to 🎩 this, since I have no access to a real store whose API responds this way. However, it passes tests and we'll have real life users testing it pretty quickly.

### Checklist:

---

### Related issues:
Closes:

#111 and #113

### DIFF
--- a/lib/shopify_transporter/exporters/magento/soap.rb
+++ b/lib/shopify_transporter/exporters/magento/soap.rb
@@ -91,7 +91,7 @@ module ShopifyTransporter
             }
           ).body
 
-          session_id = result&.dig(:login_response, :login_return)
+          session_id = result&.dig(:login_response, :login_return) || result&.dig(:login_response_param, :result)
           raise FailedLoginError, result.to_s unless session_id.present?
 
           @soap_session_id ||= session_id
--- a/spec/shopify_transporter/exporters/magento/soap_spec.rb
+++ b/spec/shopify_transporter/exporters/magento/soap_spec.rb
@@ -56,10 +56,18 @@ module ShopifyTransporter
             expect { Soap.new(init_params).call(:test_call, {}) }.to raise_error(Soap::FailedLoginError, expected_error_message)
           end
 
-          it 'creates a session correctly if login response contains a session id' do
+          it 'creates a session correctly if the login response contains a session id in format A' do
             mock_client = spy('mock_client')
             stub_client_call(mock_client)
             stub_login_call(mock_client)
+
+            Soap.new(init_params).call(:test_call, {})
+          end
+
+          it 'creates a session correctly if login response contains a session id in format B' do
+            mock_client = spy('mock_client')
+            stub_client_call(mock_client)
+            stub_login_call(mock_client, body: { login_response_param: { result: '456' } })
 
             Soap.new(init_params).call(:test_call, {})
           end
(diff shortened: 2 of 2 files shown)


spec/shopify_transporter/exporters/magento/soap_spec.rb
```diff
@@ -56,14 +56,22 @@ def stub_login_call(mock_client, body: { login_response: { login_return: '123' }
             expect { Soap.new(init_params).call(:test_call, {}) }.to raise_error(Soap::FailedLoginError, expected_error_message)
           end
 
-          it 'creates a session correctly if login response contains a session id' do
+          it 'creates a session correctly if the login response contains a session id in format A' do
```
surajreddy: minor: prefer being more explicit: `with keys :login_response and :login_return`

spec/shopify_transporter/exporters/magento/soap_spec.rb
```diff
@@ -56,14 +56,22 @@ def stub_login_call(mock_client, body: { login_response: { login_return: '123' }
...
             mock_client = spy('mock_client')
             stub_client_call(mock_client)
             stub_login_call(mock_client)
 
             Soap.new(init_params).call(:test_call, {})
           end
 
+          it 'creates a session correctly if login response contains a session id in format B' do
```
surajreddy: minor: prefer being more explicit: `with keys : login_response_param and : result `

lib/shopify_transporter/exporters/magento/soap.rb
```diff
@@ -91,7 +91,7 @@ def soap_session_id
             }
           ).body
 
-          session_id = result&.dig(:login_response, :login_return)
+          session_id = result&.dig(:login_response, :login_return) || result&.dig(:login_response_param, :result)
```
fahia: Wait, how do we know this is the end of it? This may solve the one problem but does each user's API have the chance of having some weird result? How many permutations of `login_response` can there be?

JordanVerasamy: Well, yeah. Like I said, this is _not_ a permanent solution, it's a stopgap.

fahia: But are we not investigating why this sort of thing is happening and how weird it can get? Are we comfortable shipping fixes to things we're not fully understanding?

JordanVerasamy: 100%. Like I said in the issues, I want to investigate more deeply into why this is actually happening. If it was just one user, I'd tell them to fork the code and fix it on their end while we do that. But since 2 separate users already logged the same problem, I'm assuming it's widespread and I'd rather merge a fix while we continue investigating.

fahia: 👍

Issue #109, opened 2018-10-31
Conversion does not take special_price in Magento into account.


JordanVerasamy:
#### CAUTION: Do not attach any sensitive information or PII

Please be careful not to include sensitive information in the files because this is a public repository (for example, real customer names, emails and addresses).

#### Problem:
Describe the problem you are experiencing with the gem.

Special_price is ignored.

#### Steps to Reproduce:
List the steps we can take to reproduce the problem:

1. Import a product with both price and special price
2. Special price is ignored, price in Shopify matches price in Magento

JordanVerasamy:
Update: This is not as simple as we thought.

Turns out that there's a bunch more logic that determines what price actually gets displayed in Magento.

Most importantly, Ziyan discovered that configurable products can have special prices too, not just simple products.

The parent configurable product can override the child's `price` or `special_price`, which means we have to do a check against other records every time we want to decide what to import, and that we are sensitive to which of the parent and child we see first unless we add additional logic for that.

But then if the price we're actually displaying is the parent's special price, then what should we import as the compare at price in Shopify -- the child's price? the parent's price? the child's special price? Since these are _all_ being overridden, it's not obvious what the user would want.

Also, turns out there's quite a bit of logic that governs what prices actually get displayed in Magento (there's actually an entire section of the admin devoted to [price display rules](https://docs.magento.com/m2/ce/user_guide/tax/display-settings.html)) so it's impossible anyways to import prices into Shopify that exactly match how they're displayed in Magento.

We were adding this ticket to our Q4 plans because we thought it was very easy and unambiguous. Now that we know there's some ambiguity and some room for decision making, I think that this feature should be reprioritized and either moved down or left permanently in the realm of custom pipeline stages.

PR #119, opened 2018-11-12
Convert special_price smartly to make compare_at_price work as expected

JordanVerasamy:
### What it does and why:

Converts Magento `special_price` into Shopify's `compare_at_price` properly.

### Demo

In Magento:

image

In Shopify:

image

### Checklist:

---

### Related issues:
Closes:

https://github.com/Shopify/shopify_transporter/issues/109


JordanVerasamy:
^ Turns out this is much more ambiguous than we thought, because of the fact that parent products can have prices and special prices as well and these override the child's.

PR #120, opened 2018-11-13
[R] force user to pass along scheme prefix in hostname

JordanVerasamy:
### What it does and why:

Instead of passing `magento-sandbox.myshopify.io` as the hostname, pass `https://magento-sandbox.myshopify.io` instead. This allows users to specify either `http` or `https` as their transfer protocol, for example.

### Checklist:

---

### Related issues:
Closes:


surajreddy:
I think this should be a major version bump as it will break compatibility with older config files. 

Alternatively, we could check whether the url has http[s] in it already and only append `https` if it is not already in the url. This would maintain compatibility with existing config files.

Thoughts?

PR #122, opened 2018-11-16
[R] implement backwards compatible hostname scheme reader

JordanVerasamy:
### What it does and why:

Use a regex matcher to determine what format the hostname was provided in, and massages it into the right format. Assumes `https` as default transfer protocol if none is provided.

### Checklist:

---

### Related issues:
Closes:

### DIFF
--- a/lib/shopify_transporter/exporters/magento/soap.rb
+++ b/lib/shopify_transporter/exporters/magento/soap.rb
@@ -73,11 +73,19 @@ module ShopifyTransporter
         end
 
         def soap_client
+          default_scheme = 'https://'
+          hostname_with_scheme = protocol_provided?(@hostname) ? @hostname : default_scheme + @hostname
+
           @soap_client ||= Savon.client(
-            wsdl: "#{@hostname}/api/v2_soap?wsdl",
+            wsdl: "#{hostname_with_scheme}/api/v2_soap?wsdl",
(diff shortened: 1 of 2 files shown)


fahia:
rubcop errors

surajreddy:
🎩 looks good.

lib/shopify_transporter/exporters/magento/soap.rb
```diff
@@ -74,12 +74,17 @@ def call_with_retries(method, params, retry_count = 0)
...
-            wsdl: "#{@hostname}/api/v2_soap?wsdl",
+            wsdl: "#{hostname_with_scheme}/api/v2_soap?wsdl",
             open_timeout: 500,
             read_timeout: 500,
           )
         end
 
+        def hostname_with_scheme
+          uri = URI.parse(@hostname)
+          uri.scheme.present? ? @hostname : 'https://' + @hostname
```
JordanVerasamy: I wish I could just do something like this instead:

```
def hostname_with_scheme
          uri = URI.parse(@hostname)
          uri.scheme ||= 'https'
          uri.to_s
end
```

But unfortunately that doesn't work. The URI library doesn't do the `//` thing, so you end up getting `https:example.magento.com`. I'm open to suggestions on how to get this method cleaner.

Issue #124, opened 2018-11-19
No implicit conversion of String into Integer (TypeError) when converting orders


michaelwolz:
![screenshot 2018-11-19 16 04 16](https://user-images.githubusercontent.com/7479806/48715260-e331e880-ec14-11e8-94ec-32ad1ce46091.png)

khoimm92:
Hi there, i also face this problem. Any update?

surajreddy:
Hi folks, sorry for the delay in taking a look at this! 

It seems to be a problem with extracting discounts from the order.

Does the conversion complete if you remove the discount stage step in your configuration file and change it to this?

```
  order:
    record_key: increment_id
    pipeline_stages:
      - TopLevelAttributes
      - LineItems
      - AddressesAttribute
      - Transactions
      - ShippingLines
```

(Removed `- Discounts`)

zliu1020:
Hi @michaelwolz,
I tested an order with fixed amount discounts and was unable to reproduce the bug locally. Would you mind posting the product JSON if you are comfortable in doing so?

michaelwolz:
@surajreddy Yes, if I remove the discount pipeline it works. 

@zliu1020 do you mean the product or the order JSON? We didn't import products with the Transporter App because this was all handled by our ERP System automatically. Is the product JSON needed for the conversion of the orders?

surajreddy:
Thank you for the update @michaelwolz!

You don't need the product JSON for the conversion of the orders.

It seems that your exports have discount data in a format we didn't predict. Would you be able to share a few of orders that have discounts from your JSON export with the customer information removed? (name, email, addresses)?

surajreddy:
@michaelwolz thanks! The file looks good, but I'll remove it from the comment just in case.

surajreddy:
@michaelwolz unfortunately the sample you shared is working with the discount pipeline/does not generate the error. I'll send you a private message to follow up to see if we can access to more of the extracted data to test with.

surajreddy:
@michaelwolz I sent you an email to the address on your website - let me know if you don't get it.

khoimm92:
@michaelwolz My store has some test data, and I think it caused the above error.
I had modify this function, and it's worked.
`def value_as_float(hash, key)
     begin
            hash[key].present? ? hash[key].to_f : 0
    rescue
           0
    end
end`

khoimm92:
[...]

PR #126, opened 2018-12-11
Fix No implicit conversion of String into Integer TypeError in discount pipeline stage.

zliu1020:
### What it does and why:
Properly handle the case where the `line_items` is a singular hash instead of an array.

### Checklist:

---

### Related issues:
Closes: https://github.com/Shopify/shopify_transporter/issues/124

### DIFF
--- a/lib/shopify_transporter/pipeline/magento/order/discounts.rb
+++ b/lib/shopify_transporter/pipeline/magento/order/discounts.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 require 'shopify_transporter/pipeline/stage'
 require 'shopify_transporter/shopify'
-
+require 'pry'
 module ShopifyTransporter
   module Pipeline
     module Magento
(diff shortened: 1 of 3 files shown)


lib/shopify_transporter/pipeline/magento/order/discounts.rb
```diff
@@ -82,9 +82,11 @@ def value_as_float(hash, key)
           def qualifies_for_percentage_discount?(hash)
             return false unless line_items?(hash)
 
-            return true if line_items(hash).is_a?(Hash) && value_as_float(line_items(hash), 'discount_percent') > 0
-
-            all_discount_percentages(hash).length == 1 && all_discount_percentages(hash).first > 0
+            if line_items(hash).is_a?(Hash)
```
JordanVerasamy: i don't understand how this changes any of the behaviour

JordanVerasamy:
Not really following what's going on in this PR. Could you explain in more detail?

lib/shopify_transporter/pipeline/magento/order/discounts.rb
surajreddy: We're not calling `value_as_float(line_items(hash), 'discount_percent') > 0`

JordanVerasamy:
wait now i get it

JordanVerasamy:
This change fixes the behaviour when it's a `Hash` but the value as float is negative. Right?

zliu1020:
To tophat, try running the file from Wolz again. We should no longer see any error this time.

JordanVerasamy:
Right, sorry, when I said negative I really meant non-positive. My b.

Issue #123, opened 2018-11-19
NoMethodError when converting orders


michaelwolz:
When converting orders I am getting a NoMethodError.

I disabled the discount pipeline in the config because there was an error too. 

![screenshot 2018-11-19 15 53 28](https://user-images.githubusercontent.com/7479806/48714769-b4ffd900-ec13-11e8-9206-b4602dec6dc3.png)

Thanks.

zliu1020:
Hi @michaelwolz,

Our code is making the assumption that if two products share identical SKU, one of them would be a configurable product, and the other one would be simple product associated with that configurable one. 

We made this assumption based on our sandbox data and our best understanding of how Magento works. The fact that you are getting such an error reveals that our assumption is wrong in some cases. 

Could you share the JSON file to help us debug, if that's okay with you?

michaelwolz:
@zliu1020 see my comment on #124

surajreddy:
Thank you for some of the data you sent us for this @michaelwolz! We'll take a look and get back to you soon.

surajreddy:
@michaelwolz this should be fixed up now as well - can you try your conversion again after upgrading the gem?

michaelwolz:
@surajreddy it works just fine :) Thank you.

I wish all of you a merry christmas and a happy new year :)

surajreddy:
@michaelwolz glad to hear that! Merry christmas and happy new year to you as well!

PR #128, opened 2018-12-12
Enhance the check for merging child line item into parent line item

zliu1020:
### What it does and why:

Merge the child line item into the parent line item only when:
- Parent is present
- Child is present
- Child has a price of 0

### Checklist:

---

### Related issues:
Closes: https://github.com/Shopify/shopify_transporter/issues/123

### DIFF
--- a/lib/shopify_transporter/pipeline/magento/order/line_items.rb
+++ b/lib/shopify_transporter/pipeline/magento/order/line_items.rb
@@ -26,13 +26,23 @@ module ShopifyTransporter
             line_items.group_by { |line_item| line_item['sku'] }.map do |_, associated_items|
               case associated_items.size
               when 1
-                associated_items.first
+                associated_items.first.except('product_type')
               when 2
                 parent = associated_items.find { |x| x['product_type'] == 'configurable' }
                 child = associated_items.find { |x| x['product_type'] == 'simple' }
-                parent.merge(child.slice('name'))
-              end.except('product_type')
+                return line_items_without_type(line_items) unless should_merge_child_into_parent?(parent, child)
+
+                parent.merge(child.slice('name')).except('product_type')
+              end
             end
+          end
+
+          def should_merge_child_into_parent?(parent, child)
+            parent.present? && child.present? && child['price'].present? && child['price'].to_f == 0
+          end
+
+          def line_items_without_type(line_items)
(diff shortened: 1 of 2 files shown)


JordanVerasamy:
🎩 looks like this solves @michaelwolz's issue properly.

lib/shopify_transporter/pipeline/magento/order/line_items.rb
```diff
@@ -26,13 +26,23 @@ def combine_associated_line_items(line_items)
...
+                associated_items.first.except('product_type')
               when 2
                 parent = associated_items.find { |x| x['product_type'] == 'configurable' }
                 child = associated_items.find { |x| x['product_type'] == 'simple' }
-                parent.merge(child.slice('name'))
-              end.except('product_type')
-            end
+                next line_items_without_type(associated_items) unless should_merge_child_into_parent?(parent, child)
+
+                next parent.merge(child.slice('name')).except('product_type')
```
surajreddy: We should add en else case to handle potential cases where there somehow are 3 or more line items with the same sku. I think we can still combine them if the price is `0`, but we should at least return something.

zliu1020: I added an `else` statement so that the code will return the line items without merging anything. I don't know if we should try to combine the items if any price is `0`. It's hard to figure out the merchant's original intention of more than 3 line item sharing the same SKU. Was it a mistake or was it intentional? Which line items should we merged into if there are multiple ones with different non-zero prices..? I feel like what I can do here is very limited  and cannot afford doing the wrong guess.

lib/shopify_transporter/pipeline/magento/order/line_items.rb
```diff
@@ -26,13 +26,25 @@ def combine_associated_line_items(line_items)
...
+              else
+                next line_items_without_type(associated_items)
```
surajreddy: minor: I think we don't need a next here and at the last line of `when 2`.

lib/shopify_transporter/pipeline/magento/order/line_items.rb
surajreddy: Agreed. Definitely better to have more data than wrong data for these cases, until we know for sure.

Issue #130, opened 2018-12-20
No implicit conversion of Hash into String (TypeError)


zliu1020:
Error reported by someone converting orders with addresses:
```
/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/shopify_transporter-2.5.1/lib/shopify_transporter/pipeline/magento/order/addresses_attribute.rb:47:in `+':
no implicit conversion of Hash into String (TypeError)
```
We are investing this, please add details here if you are running into the same issue.

PR #131, opened 2018-12-20
Bug fix - No implicit conversion of Hash into String (TypeError)

zliu1020:
### What it does and why:
User's file has order address information in unexpected format. 
We expect the values (name, phone, fax, company, etc) to be string in the address hash, whereas in their file the value can be a nonsense hash like `{"@xsi:type": "xsd:string"}`.

This is most likely due to their customization of Magento allowing mandatory fields (like last name) to be empty and our XML phaser did the wrong move. However, Transporter CLI should not crash on such condition. The more elegant solution is to skip the troublesome fields and let the user know skipping happened. 

This pr adds the functionality to purge the address before the conversion step happens and prints a warning informing the user some fields get skipped due to unexpected format. 

Sample warning message:
```
Warning: Order 10003210 - telephone of the billing address is in an unexpected format. Transporter CLI expects it to be a string. Skipping telephone.
Warning: Order 10003210 - fax of the billing address is in an unexpected format. Transporter CLI expects it to be a string. Skipping fax.
```

### Checklist:

---

### Related issues:
Closes: https://github.com/Shopify/shopify_transporter/issues/130

### DIFF
--- a/lib/shopify_transporter/pipeline/magento/order/addresses_attribute.rb
+++ b/lib/shopify_transporter/pipeline/magento/order/addresses_attribute.rb
@@ -23,10 +23,10 @@ module ShopifyTransporter
             result = items && items['result']
             address = result && result["#{prefix}address"]
             address ||= []
-            purge_address(input, address, prefix)
+            clean_address(input, address, prefix)
           end
 
-          def purge_address(input, address, prefix)
+          def clean_address(input, address, prefix)
             address.each_with_object({}) do |(key, val), purged_address|
               if val.is_a?(String)
                 purged_address[key] = val
@@ -59,7 +59,9 @@ module ShopifyTransporter
           end
 
           def name(address_attrs)
-            (address_attrs['firstname'].to_s + ' ' + address_attrs['lastname'].to_s).strip
+            if address_attrs['firstname'].present? && address_attrs['lastname'].present?
+              address_attrs['firstname'] + ' ' + address_attrs['lastname']
+            end
           end
         end
       end
--- a/lib/shopify_transporter/pipeline/magento/order/addresses_attribute.rb
+++ b/lib/shopify_transporter/pipeline/magento/order/addresses_attribute.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 require 'shopify_transporter/pipeline/stage'
 require 'shopify_transporter/shopify'
-require 'pry'
+
 module ShopifyTransporter
   module Pipeline
     module Magento
--- a/lib/shopify_transporter/pipeline/magento/order/addresses_attribute.rb
+++ b/lib/shopify_transporter/pipeline/magento/order/addresses_attribute.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 require 'shopify_transporter/pipeline/stage'
 require 'shopify_transporter/shopify'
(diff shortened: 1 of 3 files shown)


lib/shopify_transporter/pipeline/magento/order/addresses_attribute.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 require 'shopify_transporter/pipeline/stage'
 require 'shopify_transporter/shopify'
+
```
surajreddy: Thank you for the whitespace! ❤️

lib/shopify_transporter/pipeline/magento/order/addresses_attribute.rb
```diff
@@ -21,7 +22,22 @@ def get_nested_address(input, prefix)
             items = input['items']
             result = items && items['result']
             address = result && result["#{prefix}address"]
-            address || []
+            address ||= []
+            purge_address(input, address, prefix)
+          end
+
+          def purge_address(input, address, prefix)
```
surajreddy: I prefer a different word here than `purge` - what do you think of `clean_address` or `filter_address`? Purge makes me think that we're deleting the entire address, but we're purging nested parts of it.

lib/shopify_transporter/pipeline/magento/order/addresses_attribute.rb
```diff
@@ -21,7 +22,22 @@ def get_nested_address(input, prefix)
...
+            address.each_with_object({}) do |(key, val), purged_address|
+              if val.is_a?(String)
+                purged_address[key] = val
+              else
+                address_type = prefix.chomp('_')
+                warning = "Warning: Order #{input['increment_id']} - "\
+                  "#{key} of the #{address_type} address is in an unexpected format. "\
+                  "Transporter CLI expects it to be a string. Skipping #{key}."
```
surajreddy: This is a clear and informative message! 💯

lib/shopify_transporter/pipeline/magento/order/addresses_attribute.rb
```diff
@@ -43,9 +59,7 @@ def address_attributes(input, prefix)
           end
 
           def name(address_attrs)
-            if address_attrs['firstname'].present? && address_attrs['lastname'].present?
-              address_attrs['firstname'] + ' ' + address_attrs['lastname']
-            end
+            (address_attrs['firstname'].to_s + ' ' + address_attrs['lastname'].to_s).strip
```
surajreddy: We should leave this the same. Ideally, we shouldn't construct this if we have only first_name or only last_name.

lib/shopify_transporter/pipeline/magento/order/addresses_attribute.rb
```diff
@@ -21,7 +22,22 @@ def get_nested_address(input, prefix)
             items = input['items']
             result = items && items['result']
             address = result && result["#{prefix}address"]
-            address || []
+            address ||= []
```
mimperatore: `address ||= {}`

mimperatore: Alternatively, just `address = (result && result["#{prefix}address"]) || {}`

PR #143, opened 2019-05-21
Error trapping on transporter

SecondeJK:
returns 0 int if there is no response

### Checklist:

---

### Related issues:
Closes:

### DIFF
--- a/lib/shopify_transporter/exporters/magento/product_exporter.rb
+++ b/lib/shopify_transporter/exporters/magento/product_exporter.rb
@@ -115,6 +115,18 @@ module ShopifyTransporter
           @client
             .call(:catalog_inventory_stock_item_list, products: { product_id: product_id })
             .body[:catalog_inventory_stock_item_list_response][:result][:item][:qty].to_i
+            
+          if res.nil?
+            return 0
+          end
+              
+          item = res[:item]
+          if item.nil?
+            return 0
+          end
+              
+          qty = item[:qty].to_i
+          return qty
         end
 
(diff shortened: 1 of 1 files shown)


lib/shopify_transporter/exporters/magento/product_exporter.rb
```diff
@@ -115,6 +115,18 @@ def inventory_quantity_for(product_id)
           @client
             .call(:catalog_inventory_stock_item_list, products: { product_id: product_id })
             .body[:catalog_inventory_stock_item_list_response][:result][:item][:qty].to_i
+            
+          if res.nil?
```
mimperatore: I'm afraid this isn't going to work since `res` isn't defined.

You may want to do something like this instead:
```
res_body = @client.call(
  :catalog_inventory_stock_item_list,
  products: { product_id: product_id }
).body
res_body.dig(:catalog_inventory_stock_item_list_response,:result,:item,:qty)&.to_i || 0 
```
A test should also be added, but the current tests are not for the faint of heart and they beg to be refactored.  Perhaps there's someone who'd like to tackle that... it could be a neat little project and a great opportunity to learn more about `rspec`.

mimperatore (changes requested):

mimperatore:
Thanks for putting this together @SecondeJK!  I provided some suggestions, and also fixed the build pipeline so that it actually works now (you'll need to rebase from master before you push again).

Not sure if you were aware, but you can run `bundle exec rspec` locally to see if your code still lets all the tests pass.

